PR #3, opened 2016-06-25
Upgrade url

carols10cents:
Hi! So I'm working on upgrading the url crate in crates.io, since I knew it would be A Thing since there were breaking changes in 1.0. oauth2 was the only crate that also needed its dependency updated to get all usages of url to be > 1.0, I know it might not matter if none of the crates in crates.io do a `pub export` of url-related things, but I just went ahead and did it anyway.

I tried this branch out with a local crates.io instance and I was able to log in with github!


alexcrichton:
Thanks!

PR #5, opened 2016-10-14
update curl

steveklabnik:
I _think_ that this maps properly to the new API, but since there are no tests, I'm not 100% sure. Does this look alright? This now has more unwraps, but it seems like unwrap isn't totally forbidden here.

I also removed a public trait, because I wasn't sure what to implement it on now. 

If this looks reasonable, I'll push another commit with a version bump.


alexcrichton:
👍

PR #6, opened 2017-05-04
Allow setting the grant_type

lipanski:
hello

according to the [OAuth2 RFC](https://tools.ietf.org/html/rfc6749#section-4.1.3) the `grant_type` is mandatory when requesting a token. I'm implementing the Google OAuth2 flow and it's definitly mandatory there. hence this PR :)

it changes the API by adding a new argument to the `#exchange` method so I'd like to hear your thoughts or suggestions about this.

thanks!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,6 +8,7 @@ use url::Url;
 use std::collections::HashMap;
 
 use std::io::Read;
+use std::convert::Into;
 use curl::easy::Easy;
 
 /// Configuration of an oauth2 application.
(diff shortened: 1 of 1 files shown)


lipanski:
actually this might look better as an `enum GrantType` and maybe with a different interface that, cause the different grant types require different parameters:

```rust
fn code(code: String) {}
fn client_credentials() {}
fn password(username: String, password: String) {}
```

or prepend these with `exchange_` or `grant_`?

alexcrichton:
Awesome, thanks! The original use case here was GitHub oauth for crates.io, so I guess GitHub doesn't need it? An odd quirk!

I wonder if maybe this should be changed to a more builder-style interface? That way we could expose more configuration and perhaps add options in a backwards compatible way too

lipanski:
Added this to https://github.com/alexcrichton/oauth2-rs/pull/7 so closing this one...

PR #7, opened 2017-05-04
Parse JSON access token responses based on the content-type header and set the grant type

lipanski:
hello

the [RFC](https://tools.ietf.org/html/rfc6749#section-5.1) mentions that a successful access token response will be represented in JSON format. I didn't want to break the old implementation so I based the JSON parsing off the `content-type` header.

this PR adds `serde` to the mix but I guess there's no way around that, if we're going for the RFC. the Google OAuth2 flow I'm implementing is adhering to that and there was no way to make it produce `application/x-www-formurlencoded` (there's no magic query parameter and I've also tried the `Accept` header).

regarding the new optional `Token` fields, the [same RFC](https://tools.ietf.org/html/rfc6749#section-5.1) mentions that `access_token` and `token_type` are mandatory, while `expires_in`, `refresh_token` and `scope` are optional (and nice to have). the latter being mandatory with the current implementation so I'd actually like to turn that into an `Option` - wdyt? (in the case of the Google API, the `scope` is actually missing)

I'd also like to process the OAuth2 error fields as a separate `struct` - maybe `TokenError` and return `Result<Token, TokenError>` - but I first wanted to know what you think of my implementation.

thanks :pizza: 

---

**Update:** I've merged https://github.com/alexcrichton/oauth2-rs/pull/6 into this (but as separate commits)


alexcrichton:
Looks great to me, thanks!

What do you think of perhaps moving to a struct with private fields and public accessor functions? That'd be more API-extensible into the future which'd be nice.

I'm fine either way though, just curious what you think!

src/lib.rs
```diff
@@ -47,31 +72,66 @@ impl Config {
...
         let mut url = self.auth_url.clone();
         url.query_pairs_mut().clear().extend_pairs(
             pairs.iter().map(|&(k, v)| { (k, &v[..]) })
         );
-        return url;
+
+        url
+    }
+
+    #[deprecated(since="0.4.0", note="please use `exchange_code` instead")]
```
lipanski: not sure about what version you'll release or whether I should mark it as deprecated or remove it altogether. thoughts?

src/lib.rs
```diff
@@ -133,9 +208,23 @@ impl Config {
         if token.access_token.len() != 0 {
             Ok(token)
         } else if error.len() > 0 {
-            Err(format!("error `{}`: {}, see {}", error, error_desc, error_uri))
+            let token_error = TokenError { error, error_description, error_uri };
```
lipanski: craaaaazy :boom:

(but yeah, this means the crate will only work with the latest stable Rust - is that ok?)

lipanski:
@alexcrichton I've merged my other PR into this one and made all my changes backwards compatible. tried the thing out on my code (involving Google OAuth2) and it works.

in any case, the Github API is probably very flexible (also because it's Ruby/Rails) and I don't think you'd have problems with the `authorization_code`. JSON you probably won't get without setting an `Accept` header or something so I wouldn't touch the response part.

there's one other thing - the `response_type` request parameter is also something *mandatory* as part of the authorization URL. so far I've worked around it by just appending to the `Url` object. I'm fine with that for now, but I'd consider making another PR to add it. it can take various values, so I'd expose it through a method - maybe by using the builder pattern or just add another argument to `authorize_url` (?!). again, I doubt the Github API would complain about that (if it doesn't require it in the first place) - but I'll try it out as well.

last but not least, I think of adding a few tests in a different PR (and maybe even some docs).

would it be possible for you to add a licence though? (not one of the draconic ones :)

src/lib.rs
```diff
@@ -20,16 +24,37 @@ pub struct Config {
...
     pub token_type: String,
+    #[serde(default)]
+    pub expires_in: Option<u32>,
+    #[serde(default)]
+    pub refresh_token: Option<String>,
+}
+
+// See https://tools.ietf.org/html/rfc6749#section-5.2
+#[derive(Debug, Deserialize)]
+struct TokenError {
```
lipanski: this is entirely private, the error exposed in the API is still a `String`.

alexcrichton:
Oh oops! The license is [intended to be MIT/Apache-2.0](https://github.com/alexcrichton/oauth2-rs/blob/master/Cargo.toml#L5), I'll add the corresponding LICENSE-* files though

src/lib.rs
alexcrichton: Yeah that's ok, no worries!

src/lib.rs
alexcrichton: Yeah this is fine, thanks!

src/lib.rs
alexcrichton: I think due to the expansion of public fields in `Token` this'll need a major version bump anyway, so want to go ahead and expose this as a public type?

alexcrichton:
Looks great to me, thanks again! I left a few comments, but to reiterate I don't mind merging at any time, just let me know!

src/lib.rs
lipanski: well actually it's a bit more complicated. on one side, if I make the `TokenError` public, it won't really do much unless I also get all those `exchange` methods return `Result<Token, TokenError>`. however, if I do the latter, I can't return the (json) parse errors any more, cause I'd have to fit them somehow in the `TokenError` pattern.

regarding the `TokenError`, the `error` field should/could be an `Enum` (the [RFC](https://tools.ietf.org/html/rfc6749#section-4.1.2.1) mentions a strict set of values) - but again this depends on some investigation that I'll have to do whether popular APIs (let's say Google, Github, Twitter) honour this.

so please go ahead with merging this, and I'll follow up with some more stuff (including tests) and some questions :)

personally I don't think that adding extra fields to the `Token` struct would be considered a breaking change - I doubt there's anyone initializing the `Token` struct directly (and not through the `exchange` method) and if that holds, it shouldn't conflict with existing implementations. or am I missing something?

alexcrichton: [crates.io is](https://github.com/rust-lang/crates.io/blob/9ec474cc62e77c35a88fe25a31cbb44d27a5cc72/src/http.rs#L70-L78) :(

alexcrichton:
Ok, let's merge.

src/lib.rs
lipanski: oh, didn't think of that.. but it's fixable :champagne:

PR #9, opened 2017-05-10
Show error message on failure

ivegotasthma:
I'm tryring out the oauth2 crate and while learning the flow of it it's nice to see the error messages returned from the server.


alexcrichton:
Thanks! Perhaps this could use `from_utf8_lossy` to avoid the panics?

ivegotasthma:
@alexcrichton Thanks for the advice, I adjusted it.

alexcrichton:
Thanks!

PR #19, opened 2017-08-10
don't clear params

KeenS:
some authorization URL requires additional params like `https://example.com/login?username=someuser&...` so I'd like to preserve params included in the authorization uri.
Someone may want additional APIs like `add_param`. I'll implement it if you want


lipanski:
:+1: no, looks ok like that. I'll release a new version.

lipanski:
1.1.2 is out :car:

PR #24, opened 2018-01-09
Allow configuring basic-auth usage

### LINKED ISSUE #22: Support for basic-auth to exchange client credentials
Currently, this library always includes `client_id` and `client_secret` as form parameters in the request body.

This is not in-line with what the RFC requires servers to expect, and thus this library doesn't work with some servers (which I have encountered in the wild).

The specific section of [RFC 6749](https://tools.ietf.org/html/rfc6749) I'm referencing is [2.3.1](https://tools.ietf.org/html/rfc6749#section-2.3.1) which says:

> The authorization server MUST support the HTTP Basic authentication scheme for authenticating clients that were issued a client password.
> ...
> Alternatively, the authorization server MAY support including the client credentials in the request-body

This library should either support, or default to, sending client credentials using basic-auth.

For reference, here's what a few other common oauth libraries do:

* golang/oauth2 -- [golang/oauth2](https://github.com/golang/oauth2/blob/876b1c/internal/token.go#L93-L170) library always uses basic auth, unless the domain is in a whitelist of "broken implementations", which notably includes google.
* ruby oauth2 -- [Ruby oauth2](https://github.com/oauth-xx/oauth2/blob/v1.4.0/lib/oauth2/client.rb#L38) provides an option which defaults to request body.
* node simple-oauth2 -- [simple-oauth2](https://github.com/lelylan/simple-oauth2/blob/1.5.0/README.md#options) provides the `useBasicAuth` option, which defaults to false.

Since this library is fairly low level, I think supporting providing an option (as most other libraries support) is the right thing to do. I'm willing to implement that if it sounds reasonable.

euank:
The RFC specifically recommends using basic-auth for client credential
grants (https://tools.ietf.org/html/rfc6749#section-2.3).

Some oauth2 servers also require using basic auth for such requests.

fixes #22

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -112,11 +112,25 @@ pub struct Config {
     client_id: String,
     client_secret: String,
     auth_url: Url,
+    auth_type: AuthType,
     token_url: Url,
     scopes: Vec<String>,
     response_type: ResponseType,
     redirect_url: Option<String>,
     state: Option<String>,
+}
+
+/// 
+/// Indicates whether requests to the authorization server should use basic authentication or
+/// include the parameters in the request body for requests in which eithr is valid.
+///
+/// The default AuthType is *RequestBody*
+pub enum AuthType {
+    /// The client_id and client_secret will be included as part of the request body.
+    RequestBody,
+    /// The client_id and client_secret will be included using the basic auth authentication
+    /// scheme.
+    BasicAuth,
 }
 
 impl Config {
@@ -130,6 +144,7 @@ impl Config {
             client_id: client_id.into(),
             client_secret: client_secret.into(),
             auth_url: Url::parse(auth_url.as_ref()).unwrap(),
+            auth_type: AuthType::RequestBody,
             token_url: Url::parse(token_url.as_ref()).unwrap(),
             scopes: Vec::new(),
             response_type: ResponseType::Code,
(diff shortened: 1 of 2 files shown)


euank:
ping @lipanski; am I missing something I should do to progress this PR?

src/lib.rs
```diff
@@ -112,13 +112,27 @@ pub struct Config {
...
     token_url: Url,
     scopes: Vec<String>,
     response_type: ResponseType,
     redirect_url: Option<String>,
     state: Option<String>,
 }
 
+/// 
+/// Indicates whether requests to the authorization server should use basic authentication or
+/// include the parameters in the request body for requests in which eithr is valid.
```
lipanski: *either* (typo)

src/lib.rs
```diff
@@ -112,13 +112,27 @@ pub struct Config {
...
+///
+/// The default AuthType is *RequestBody*
```
lipanski: please add another line with `///` after this (like all other block comments)

src/lib.rs
```diff
@@ -160,6 +175,19 @@ impl Config {
         self
     }
 
+    ///
+    /// Allows configuring whether basic auth is used to communicate with the authorization server
+    /// or not. This option affects 
+    /// auth authentication scheme.
```
lipanski: please move this to the line above - it fits just fine :)

lipanski:
I'll just merge and fix those style issues myself cause it's easier. thanks for this & sorry for the delay :1st_place_medal:

euank:
Thanks!

PR #25, opened 2018-01-09
Include scope in exchange_client_credentials

### LINKED ISSUE #23: client_credentials requests should also include `scope`
As far as I can tell from reading the RFC, it's expected that scope is also included for client credential type grants.

Regardless, I know of at least one server which requires the scope parameter, including when exchanging client credentials.

euank:
In practice, some servers require this, and I think this is what the RFC
intends an oauth client to do.

Fixes #23


lipanski:
thanks :cake:

lipanski:
@euank [1.2.0](https://github.com/alexcrichton/oauth2-rs/releases/tag/1.2.0) is out :car:

PR #26, opened 2018-01-31
Try to parse errors even on non-200 statuses

andor44:
In chapter 5.2 of the spec it says servers should return HTTP 400. Currently everything non-200 is interpreted as an unknown error. I'm working with [dex](https://github.com/coreos/dex/) at the moment, which will correctly return HTTP 400 when it is given an expired code, and the error is not deserialized. I just copied over the error deserialization from the token construction.


lipanski:
thanks :+1:

Issue #28, opened 2018-02-25
The "state" parameter should be per-request for CSRF protection


jtgeibel:
I'm working to update the crates.io codebase from 0.3 to 1.0+.  After updating the relevant call sites, I'm running into a borrow error because the `set_state` function is part of the builder pattern for `Config` (and takes self by value).  Previously a per-request state value was provided as a parameter to `authorize_url`.

Here is the relevant code and the diff addressing the API change:

```diff
    let state: String = thread_rng().gen_ascii_chars().take(16).collect();
     req.session()
         .insert("github_oauth_state".to_string(), state.clone());
 
-    let url = req.app().github.authorize_url(state.clone());
+    let url = req.app().github.set_state(&state[..]).authorize_url();
```

Basically, `req.app()` returns a `&Arc<App>` where `App` is a struct initialized at boot and containing a field `github` of type `oauth2::Config`.  This code results in an error "cannot move out of borrowed content".

The old API allowed a per-request state value to be easily supplied to each `authorize_url` call.  As seen in the code above this random state is stored in the session.  It is recovered and checked by the callback endpoint later in the authorization flow.

With the new API the state parameters is captured as part of the builder pattern and it seems this encourages reuse of the same state across multiple requests to the authorization server.  Since the authorization URL (and embedded state parameter) are shared with the user's browser, it is easy for anyone to predict the state if it is reused.  From section [10.2](https://tools.ietf.org/html/rfc6749#section-10.12) of the spec, where "binding value" refers to the state parameter:

> The binding value used for CSRF protection MUST contain a non-guessable value (as described in
Section 10.10), and the user-agent's authenticated state (e.g., session cookie, HTML5 local storage) MUST be kept in a location accessible only to the client and the user-agent (i.e., protected by same-origin policy).

For now, I can work around this with the following, but it seems that the current API doesn't encourage sufficient CSRF protections.

```diff
-    let url = req.app().github.authorize_url(state.clone());
+    let mut url = req.app().github.authorize_url();
+    url.query_pairs_mut().append_pair("state", &state[..]);
```

lipanski:
fair point. bit late to go back to the old interface of `authorize_url`, though that would be my preferred solution. another option would be to clear the `state` attribute within `authorize_url`, but that would be a bit surprising. any other way to keep backwards compatibility of (at least) the `authorize_url` definition?

lipanski:
[...]

Issue #33, opened 2018-03-18
Rationale for supporting `application/x-www-form-urlencoded` response


ramosbugs:
It looks like the `request_token` function supports both a JSON response and an `application/x-www-form-urlencoded` response. From my reading of the spec (https://tools.ietf.org/html/rfc6749), response bodies are always JSON:
> The parameters are included in the entity-body of the HTTP response
   using the "application/json" media type as defined by [RFC4627].  The
   parameters are serialized into a JavaScript Object Notation (JSON)
   structure by adding each parameter at the highest structure level.
   Parameter names and string values are included as JSON strings.
   Numerical values are included as JSON numbers.  The order of
   parameters does not matter and can vary.

It looks like OAuth 1.0 (https://tools.ietf.org/html/rfc5849) used `application/x-www-form-urlencoded` responses. Since this library only supports OAuth 2.0, do we need to support this response type? Technically, that's a deviation from the spec.

It might be worth removing this in the next major version since it'll break backward compatibility if anyone happens to be relying on this behavior. I'll be proposing a change soon (for #27) that would require the major version bump, so we might as well lump this in unless there's a good reason to keep it.

ramosbugs:
It looks like this was due to GitHub's early implementation of OAuth2, before the spec was finalized (https://twitter.com/patricktoomey/status/976873772601966593). As long as the client adds the `Accept: application/json` header, it should be safe to remove support for parsing form-encoded token responses. That will simplify the code significantly. Planning to roll this into upcoming 2.0.0 breaking changes. Let me know of anyone objects.

SorenHolstHansen:
Hi, sorry for using an old issue, but Slack is using application/x-www-form-urlencoded for their code exchange. I am for the moment just doing the code exchange request myself with reqwest, but would it be possible to support this again?

ramosbugs:
Hey @SorenHolstHansen,

Would you mind linking me to the relevant Slack docs? I couldn't find anything that said they require `application/x-www-form-urlencoded`.

As a workaround, I'd suggest writing a shim function to pass to the `request`/`request_async` methods that converts the `application/x-www-form-urlencoded` response to a JSON one before passing the `HttpResponse` back to this crate. There doesn't seem to be wide enough usage or anything in the published standards to justify adding this functionality back to the crate itself.

Hope that helps!

SorenHolstHansen:
[...]

Issue #27, opened 2018-02-01
Arbitrary parameters in requests and responses.


andor44:
In 8.2 the spec allows for adding arbitrary, vendor-specific (or as described in 11.2, getting them registered) parameters for both requests and responses. As an example, in my case I'm working with OIDC, which is built on top of OAuth2, which makes use of this by using an extra `id_token` field in the response for a JWT token. Would providing support for extra fields (not OIDC support itself) be in the scope of this library?

In Go everything unknown is stored into an [untyped field](https://github.com/golang/oauth2/blob/a032972e28060ca4f5644acffae3dfc268cc09db/token.go#L54) that the user can [retrieve](https://github.com/golang/oauth2/blob/a032972e28060ca4f5644acffae3dfc268cc09db/token.go#L96).

I made a quick and dirty implementation of the above approach in [here](https://github.com/andor44/oauth2-rs/tree/extra_token_fields) which works for me, but this doesn't yet support *truly* arbitrary fields, only string fields. If there's interest in it here I can flesh this out.

ramosbugs:
@andor44: I'm looking at integrating with OIDC in Rust as well, so I'd definitely be interested. Since Rust is all about static typing, I wonder if it would be possible to use generics to solve this problem. Perhaps `Token` could become a `trait` with the minimum required fields, and callers can supply their own implementations of that trait with additional fields relevant to their use case? I'm not sure if serde would support anything like this.

@alexcrichton: would this be something you'd be interested in? I'm happy to help with the effort. Some form of [OIDC](http://openid.net/specs/openid-connect-core-1_0.html) would be great, and hard-coding the 8 new request parameters and `id_token` response parameter might not be ideal (especially if there end up being conflicting standards in the future).

alexcrichton:
@ramosbugs sorry I don't quite have time any more to maintain this library, but if you'd like to pick it up I'd be more than willing to add you and/or transfer it!

ramosbugs:
Sure, if you wouldn't mind adding me, that would be great! I'd rather add OIDC functionality to this lib so others can use it, rather than building everything from scratch just for my use case. I don't see a need to transfer the whole repo though (at least at this point).

Who would be the best person to review these changes? Even if I have merge access, it would probably be best for someone else to take a look at major changes.

ramosbugs:
[...]

PR #34, opened 2018-03-29
Add authorize_url_with_state

jtgeibel:
As discussed in #28, this adds an `authorize_url_with_state` method which takes ownership of a string to use as the state parameter.  I've deprecated `set_state` and bumped the version to 1.3.1.

I've also updated the examples to generate a random state.  I can back that out or put it in a separate PR if that is preferred.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "oauth2"
 authors = ["Alex Crichton <alex@alexcrichton.com>", "Florin Lipan <florinlipan@gmail.com>"]
-version = "1.3.0"
+version = "1.3.1"
 license = "MIT/Apache-2.0"
 description = "Bindings for exchanging OAuth 2 tokens"
 repository = "https://github.com/alexcrichton/oauth2-rs"
@@ -16,3 +16,4 @@ serde_derive = "1.0"
 
 [dev-dependencies]
 mockito = "0.8.2"
+rand = "0.4"
--- a/examples/github.rs
+++ b/examples/github.rs
@@ -15,12 +15,14 @@
 
 extern crate url;
 extern crate oauth2;
+extern crate rand;
 
 use std::env;
 use std::net::TcpListener;
 use std::io::{BufRead, BufReader, Write};
 use url::Url;
 use oauth2::Config;
+use rand::{thread_rng, Rng};
 
 fn main() {
     let github_client_id = env::var("GITHUB_CLIENT_ID").expect("Missing the GITHUB_CLIENT_ID environment variable.");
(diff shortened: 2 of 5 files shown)


ramosbugs:
Thanks for the PR! I've been working on a sizable 2.0.0 rewrite in https://github.com/ramosbugs/oauth2-rs/tree/token_ext_trait that addresses #27, #28, and #33. That change (since it's allowed to be breaking for 2.x) adds a mandatory `state` argument to each `authorize_url` call, and it adds a `mod insecure` people can use if they want to ignore the advice and call `authorize_url` without a CSRF token (since technically the OAuth2 spec doesn't require it).

I think there's a good chance I'll be able to send the full rewrite out for review and merge it into https://github.com/alexcrichton/oauth2-rs/tree/devel/2.0.0 in the next ~2 weeks. Would your preference be to get this PR into an earlier release (I think it would actually need to be 1.4.0 since it adds functionality), or to wait for the 2.0.0 branch?

jtgeibel:
> That change (since it's allowed to be breaking for 2.x) adds a mandatory state argument to each authorize_url call, and it adds a mod insecure people can use if they want to ignore the advice and call authorize_url without a CSRF token (since technically the OAuth2 spec doesn't require it).

That sounds great to me @ramosbugs!  In that case, I'm closing this PR.  Once there is a 2.0 release, or maybe even a beta, we can publish an update to the 1.x series with the deprecation to alert users who are accidentally reusing random state.

I have a [commit](https://github.com/rust-lang/crates.io/commit/ce69cb20172d17fdad7899ecb73d2bd0f6f2a85b) upgrading from 1.3.0 to your branch.  The port was fairly straightforward.  I momentarily ran into some confusion using `Token` (the trait) and `basic::Token` (a type) in the same scope.  It may help if the latter was `basic::BasicToken`.  (In the end, the original code was trying to instantiate a concrete token just to pass around a string.  Now that those fields are private it was easy to factor the whole thing away, but I could see running into this in other scenarios.)

I'll probably migrate crates.io to use 2.0 once an alpha or beta package is published to... crates.io.

ramosbugs:
Thank you for dogfooding my branch and seeing what it takes to update the crates.io dependency! This saved me quite a bit of work :-)

That's great feedback about the implementation naming. I was trying to avoid the redundancy of `basic::BasicToken`, but I keep forgetting that Rust requires importing the `Token` trait in order to access the methods provided by the trait. I'll go ahead and rename everything in the `basic` module to avoid shadowing the trait names.

Other than that, does the new client interface seem reasonable?

Lastly, if you find some time to send a PR for adding the deprecation warning to the [support/1.x branch](https://github.com/alexcrichton/oauth2-rs/tree/support/1.x), that would be hugely appreciated. Otherwise, I can take care of it.

jtgeibel:
> Other than that, does the new client interface seem reasonable?

Yes, the other changes I encountered seem reasonable.  I like that the token fields are now private and the renaming of Config to Client makes sense.

> Lastly, if you find some time to send a PR for adding the deprecation warning to the support/1.x branch, that would be hugely appreciated.

I've opened #37 against the support branch.

ramosbugs:
After thinking about this a bit more, I'm not sure it makes sense to take ownership of the `state`, since the client needs to keep it around in order to verify that it matches the `state` parameter passed in via the redirect URI. Any objections to passing in a reference instead?

PR #35, opened 2018-03-30
Overhaul library to support extensibility and better adhere to RFC 6749

### LINKED ISSUE #27: Arbitrary parameters in requests and responses.

### LINKED ISSUE #28: The "state" parameter should be per-request for CSRF protection

### LINKED ISSUE #33: Rationale for supporting `application/x-www-form-urlencoded` response

ramosbugs:
NOTE: This is a breaking change that will require a major version bump.
As such, this diff bumps the next release version to 2.0.0-alpha.

This diff is a significant rewrite of the library that:
* Adds extensibility to support arbitrary OAuth 2.0 extensions in the
  future without having to break backward compatibility again. This is
  primarily accomplished through traits and generics (Resolves #27).

* Improves adherence to the OAuth 2.0 spec (RFC 6749):
  - Requires access token responses to be JSON-encoded, dropping support
    for form-encoded responses. Also adds the HTTP request header
    "Accept: application/json" to ensure that loosely compliant OAuth2
    implementations such as GitHub's return a JSON response (Resolves #33).
  - Fixes the access token error response types enum to follow Section
    5.2 of the spec instead of 4.2.2.1, which is for the Implicit Grant
    flow only. The Implicit Grant errors are only seen by the browser,
    and not by client applications that would be using this library.
  - Fixes parsing of the "scope" token response field, which is a
    space-delimited string instead of a JSON array of strings.
  - Removes the "state" field from the token error response, which is not
    defined in Section 5.2 of the spec. This field is only returned by
    earlier steps in the OAuth2 flows.
  - Only parses the access token response as an error if the client
    receives a non-200 HTTP response code. Section 5.2 of the spec defines
    the expected response code as 400 Bad Request.
  - Uses HTTP Basic Auth for client authentication by default, rather
    than including the client_id and client_secret in the request body.
    This matches the recommendation in Section 2.3.1 of the spec.

* Slightly improves the usability of the library
  - Renames the main struct from Config to Client.
  - Makes the "state" parameter mandatory unless the "insecure" module is
    used (Fixes #28). This is critical to protect clients against CSRF.
  - Improves the documentation to more closely refer to the spec.
  - Improves the examples to include CSRF protection.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /target
 /Cargo.lock
+*~
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,12 +7,10 @@ sudo: false
 branches:
   only:
     - master
-    - token_ext_trait
 script:
   - cargo test -- --test-threads=1
   - cargo test --example google
   - cargo test --example github
-  - cargo test --doc
 notifications:
   email:
     on_success: never
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,12 @@ sudo: false
 branches:
   only:
     - master
+    - token_ext_trait
 script:
   - cargo test -- --test-threads=1
   - cargo test --example google
   - cargo test --example github
+  - cargo test --doc
 notifications:
   email:
     on_success: never
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,18 +1,22 @@
 [package]
 name = "oauth2"
 authors = ["Alex Crichton <alex@alexcrichton.com>", "Florin Lipan <florinlipan@gmail.com>"]
-version = "1.3.0"
+version = "2.0.0-alpha"
 license = "MIT/Apache-2.0"
 description = "Bindings for exchanging OAuth 2 tokens"
 repository = "https://github.com/alexcrichton/oauth2-rs"
 
 [dependencies]
 curl = "0.4.0"
-url = "1.0"
+failure = "0.1"
+failure_derive = "0.1"
 log = "0.3"
 serde = "1.0"
 serde_json = "1.0"
 serde_derive = "1.0"
+url = "1.0"
 
 [dev-dependencies]
+base64 = "0.9"
 mockito = "0.8.2"
+rand = "0.4"
--- a/examples/github.rs
+++ b/examples/github.rs
@@ -13,14 +13,18 @@
 //! ...and follow the instructions.
 //!
 
-extern crate url;
+extern crate base64;
 extern crate oauth2;
+extern crate rand;
+extern crate url;
 
+use oauth2::basic::BasicClient;
+use rand::{thread_rng, Rng};
 use std::env;
 use std::net::TcpListener;
 use std::io::{BufRead, BufReader, Write};
+use std::process;
 use url::Url;
-use oauth2::Config;
 
 fn main() {
     let github_client_id = env::var("GITHUB_CLIENT_ID").expect("Missing the GITHUB_CLIENT_ID environment variable.");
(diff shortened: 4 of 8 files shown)


ramosbugs:
Once this merges, I'll send out another PR for https://github.com/ramosbugs/oauth2-rs/tree/stronger_types, which replaces stringly typed fields with stronger typing.

jtgeibel:
From the other thread:

> After thinking about this a bit more, I'm not sure it makes sense to take ownership of the state, since the client needs to keep it around in order to verify that it matches the state parameter passed in via the redirect URI. Any objections to passing in a reference instead?

Yeah, that makes sense to me.  That would eliminate a call to clone in the code using the API and it is possible that a future implementation could avoid the allocation.

I'm not sure it's worth the trouble, but an alternative API might be to take a closure that generates the random state and then returns a `(url, state)` tuple.  It would be more obvious that the consuming code is still expected to do something with it.  Or maybe the closure could be part of the `Client` struct so that the state generation code is defined along with the other configuration.

ramosbugs:
Great idea! Since I added a `CsrfToken` type in the [stronger_types](https://github.com/ramosbugs/oauth2-rs/tree/stronger_types) branch, I've been thinking of adding a factory method that creates a random one, but I don't think that should be the only interface since some callers may want to use a hash or some other value. So, a closure would work well here. I think it would make sense to leave it as an argument to `authorize_url` to help cases where people want to pass their own value. Otherwise, they'd have to update the `Client` each time, which was the issue with the 1.x implementation.

Then, I can mark `CsrfToken` as `#[must_use]` and return it from `authorize_url`as you suggested. I'll go ahead and add that to the second branch.

PR #38, opened 2018-04-15
Add stronger typing using the NewType pattern

### LINKED ISSUE #28: The "state" parameter should be per-request for CSRF protection

ramosbugs:
NOTE: This is another breaking change that will be part of the 2.0.0 release.

This diff replaces weakly typed `String`s, `Url`s, and other types with new
types generated using the `NewType` pattern. Using stronger types here
should avoid common mistakes (e.g., switching the order of the authorization
and endpoint URLs when instantiating a new `Client`).

In addition to adding a `NewType` trait, this diff adds a `NewSecretType`
trait, which implements `Debug` in a way that redacts the secret. This
behavior avoids a common source of security bugs: logging secrets,
especially when errors occur. Unlike the `NewType` trait, the
`NewSecretType` does not implement `Deref`. Instead, the secret must
be explicitly extracted by calling the `secret` method.

Finally, this PR resolves #28 by having the `authorize_url` method accept a
closure for generating a fresh CSRF token on each invocation. The token is
returned by the method as `#[must_use]`, which the caller should compare
against the response sent by the authorization server to the redirect URI.
Note that `#[must_use]` currently has no effect in this context, but it should
once https://github.com/rust-lang/rust/issues/39524 is resolved.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -533,6 +533,7 @@ new_secret_type![
     /// Client password issued to the client during the registration process described by
     /// [Section 2.2](https://tools.ietf.org/html/rfc6749#section-2.2).
     ///
+    #[derive(Deserialize, Serialize)]
     ClientSecret(String)
 ];
 new_secret_type![


ramosbugs:
Closing this PR since it points to an orphaned branch after the repo was transferred to me. Reopening separately.

PR #40, opened 2018-05-04
Using `FnOnce` for state_fn

idubrov:
Making `state_fn` to be `FnOnce` callback, so it can pass ownership of state value (for example, if provided as `|| some_precreated_csrf`)

P.S. I think, it should also make it a little bit clearer that you are supposed to only used the value once.

So instead of:
```
let (auth_url, _) = cfg.authorize_url(|| state.clone() /* state is cloned! */);
```
one could do:
```
let (auth_url, _) = cfg.authorize_url(|| state /* state is moved in! */);
```


ramosbugs:
Thanks for the PR!

PR #41, opened 2018-06-04
Make BasicAuth url-encode id and password

DeeUnderscore:
[Per RFC 6749](https://tools.ietf.org/html/rfc6749#section-2.3.1), `client_id` and `client_secret` are supposed to be url-encoded before they are used as user/pass for HTTP Basic authentication. 

I also modified a single test for this, since I didn't want to rewrite a whole bunch of them. Let me know if that's not okay.


ramosbugs:
Thanks for the PR!

PR #43, opened 2018-09-01
PKCE implementation

jongiddy:
Implementation of RFC 7636 Proof Key for Code Exchange by OAuth Public Clients.

To allow a client to be reused with unique PKCE codes, the `code_verifier` must be passed to new functions `authorize_url_pkce` and `exchange_code_pkce`.

The implementation only supports SHA256, the required default (and the most secure option).

I have also added an example of using PKCE with Microsoft Graph.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ rand = "0.4"
 serde = "1.0"
 serde_json = "1.0"
 serde_derive = "1.0"
+sha2 = "0.7"
 url = "1.0"
 
 [dev-dependencies]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -864,6 +864,31 @@ impl<EF: ExtraTokenFields, TT: TokenType, TE: ErrorResponseType> Client<EF, TT, 
     }
 
     ///
+    /// Exchanges a code produced by a successful authorization process with an access token.
+    ///
+    /// Acquires ownership of the `code` because authorization codes may only be used to retrieve
+    /// an access token from the authorization server.
+    ///
+    /// See https://tools.ietf.org/html/rfc6749#section-4.1.3
+    ///
+    pub fn exchange_code_extension(
+        &self,
+        code: AuthorizationCode,
+        extra_params: &[(&str, &str)]
+    ) -> Result<TokenResponse<EF, TT>, RequestTokenError<TE>> {
+        // Make Clippy happy since we're intentionally taking ownership.
+        let code_owned = code;
+        let mut params = vec![
+            ("grant_type", "authorization_code"),
+            ("code", code_owned.secret())
+        ];
+
+        params.extend(extra_params.iter().cloned());
+
+        self.request_token(params)
+    }
+
+    ///
     /// Requests an access token for the *password* grant type.
     ///
     /// See https://tools.ietf.org/html/rfc6749#section-4.3.2
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -220,6 +220,7 @@ extern crate serde_json;
 extern crate sha2;
 extern crate url;
 
+use std::convert::Into;
 use std::io::Read;
 use std::fmt::{Debug, Display, Error as FormatterError, Formatter};
 use std::marker::{Send, Sync, PhantomData};
@@ -377,6 +378,11 @@ macro_rules! new_type {
                 &self.0
             }
         }
+        impl Into<$type> for $name {
+            fn into(self) -> $type {
+                self.0
+            }
+        }
     }
 }
 
@@ -527,6 +533,22 @@ new_type![
 impl AsRef<str> for Scope {
     fn as_ref(&self) -> &str { self }
 }
+new_type![
+    ///
+    /// Code Challenge used for [PKCE]((https://tools.ietf.org/html/rfc7636)) protection via the
+    /// `code_challenge` parameter.
+    ///
+    #[derive(Deserialize, Serialize)]
+    PkceCodeChallengeS256(String)
+];
+new_type![
+    ///
+    /// Code Challenge Method used for [PKCE]((https://tools.ietf.org/html/rfc7636)) protection
+    /// via the `code_challenge_method` parameter.
+    ///
+    #[derive(Deserialize, Serialize)]
+    PkceCodeChallengeMethod(String)
+];
 
 new_secret_type![
     ///
@@ -570,13 +592,14 @@ new_secret_type![
     /// maximum length of 128 characters.  Each character must be ASCII alphanumeric or one of
     /// the characters "-" / "." / "_" / "~".
     ///
-    CodeVerifierS256(String)
+    #[derive(Deserialize, Serialize)]
+    PkceCodeVerifierS256(String)
     impl {
         ///
         /// Generate a new random, base64-encoded code verifier.
         ///
         pub fn new_random() -> Self {
-            CodeVerifierS256::new_random_len(32)
+            PkceCodeVerifierS256::new_random_len(32)
         }
         ///
         /// Generate a new random, base64-encoded code verifier.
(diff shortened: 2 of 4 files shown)


ramosbugs:
Thanks for the PR! I'm planning to look this over on Sunday. Apologies for the delay.

src/lib.rs
```diff
@@ -686,9 +724,9 @@ impl<EF: ExtraTokenFields, TT: TokenType, TE: ErrorResponseType> Client<EF, TT,
     ///
     /// # Arguments
     ///
-    /// * `state` - An opaque value used by the client to maintain state between the request and
-    ///   callback. The authorization server includes this value when redirecting the user-agent
-    ///   back to the client.
+    /// * `state_fn` - A function that returns an opaque value used by the client to maintain state
```
ramosbugs: nice catch! thanks for fixing these

src/lib.rs
```diff
@@ -705,15 +743,55 @@ impl<EF: ExtraTokenFields, TT: TokenType, TE: ErrorResponseType> Client<EF, TT,
...
+    where
+        F: FnOnce() -> CsrfToken,
+        G: FnOnce() -> CodeVerifier
+    {
+        let state = state_fn();
+        let verifier = verifier_fn();
+        let digest = Sha256::digest(verifier.secret().as_bytes());
+        let code_challenge = base64::encode_config(&digest, base64::URL_SAFE_NO_PAD);
+        let params = vec![
+            ("code_challenge_method", "S256"),
```
ramosbugs: It might be nice to support other methods to allow for future extensibility, but it's probably not worth complicating the API. I'm guessing S256 will remain the dominant method. In any case, callers can use `authorize_url_extension` if they want to use a different method. I agree with not supporting `plain` at all here.

Cargo.toml
```diff
@@ -14,6 +14,7 @@ rand = "0.4"
 serde = "1.0"
 serde_json = "1.0"
 serde_derive = "1.0"
+sha2 = "0.7.1"
```
ramosbugs: let's pin this to `"0.7"` so that we can pull in bug fixes without needing to push out a new release of this crate each time

tests/lib.rs
```diff
@@ -47,6 +47,30 @@ fn new_mock_client_with_unsafe_chars() -> BasicClient {
     )
 }
 
+#[test]
```
ramosbugs: nice tests! mind adding a simple one for `exchange_code_pkce` (maybe similar to `test_exchange_code_successful_with_minimal_json_response`)?

jongiddy:
Having thought about your comment about not supporting protocols other than `S256` and also about composing this with other extensions, I have doubts whether this should be merged into the library.

PKCE is just one of many extensions that OAuth2 can support. OIDC is another example, but there are also the various authorizer-specific extensions.

When there is another compelling extension that would require `authorize_url_xyz` and/or `exchange_code_xyz`, then there is likely to be someone who wants PKCE and XYZ together. Having to reimplement one or both of them outside the library to get the correct parameters would be wasteful and error-prone.

Adding `exchange_code_extension` as suggested in https://github.com/ramosbugs/oauth2-rs/issues/27#issuecomment-417834171 would allow parameters to be composed in a higher-level library.

For example:
```rust
let session =
    AuthorizationCodeFlow()
        .withCsrf(Csrf::new_random)
        .withPkce(CodeVerifier::new_random);
// build a chain of handlers

let url = session.authorize_url();
// go through chain, adding parameters
// AuthorizationCodeFlow adds "response_type"
// withCsrf adds "state"
// withPkce adds "code_challenge" and "code_challenge_type"
...
let response = <response from authorization server>

let tokens = session.exchange_code(response);
// go through chain, verifying response and adding parameters
// withCsrf verifies response "state"
// withPkce adds "verify_code"
// (probably should run in reverse, so e.g. state can be used for both CSRF and user-state)
```

Replacing `AuthorizationCodeFlow` with, for example, `OidcAuthenticationFlow` can then benefit from CSRF protection and PKCE with no other changes.

Anyway, the point is that the simpler change of adding an `exchange_code_extension` function would allow us to experiment with such composable abstractions.

Happy to hear your thoughts, even if it's no, let's just support PKCE here, in which case, I can make the changes suggested.

ramosbugs:
This brings up a lot of compelling points! I think you're right that in theory clients of this library could end up wanting arbitrary combinations of OAuth2 extensions, and having dedicated `authorize_url_*` and `exchange_code_*` functions for each of these combinations clearly wouldn't be the right approach.

I fully support adding an `exchange_code_extension` function to let clients add arbitrary extensions, whether or not we end up merging PKCE into the library.

As far as whether a particular extension should become part of this library, two criteria come to mind:
1) the complexity of the extension
2) whether it's like to be of interest to a large fraction of library clients

I think PKCE is a great candidate to include because it's both simple and likely to be used by many clients (especially native/public OAuth apps). By contrast, OpenID Connect is much more complex than the baseline OAuth2 spec, and [openidconnect-rs](https://github.com/ramosbugs/openidconnect-rs) is already a much larger library than this one. Given these criteria, I think we should focus our discussion on how best to integrate PKCE and potential future extensions into this library.

One downside I see to the "session" builder pattern above is that it works best when the same service/process will be making the calls to `authorize_url` and `exchange_code`. In many (potentially most) deployments, it's likely that the CSRF token and PKCE code verifier will need to be serialized and persisted somewhere, and then read back in when it's time to verify the CSRF token and exchange the authorization code for the access token. To support this, the session would need to be serializable/deserializable, and probably deserialize in a backward-compatible manner since a library update might be deployed in between the two steps in the OAuth flow.

Tweaking this approach slightly, we could have something like (`user_db` and `user` are just for illustrative purposes):
```rust
fn init_oauth_flow(user_db: UserDatabase, user: User) -> Url {
    let client = Client::new(client_id, client_secret, auth_url, token_url);
    let params =
        // CSRF protection is mandatory here
        AuthorizationParams::new(Csrf::new_random)
            .withPkce(CodeVerifier::new_random);
    let extensions = ...;
    let authorization_request: AuthorizationRequest =
        client.authorize_url_extension(params, extensions);

    user_db.set_user_oauth_state(
        user,
        authorization_request.csrf_token(),
        authorization_request.pkce_code_verifier(),
    );
    return authorization_request.url();
}

fn complete_oauth_flow(
    user_db: UserDatabase,
    user: User,
    code: AuthorizationCode,
    state: CsrfToken,
) -> Result<TokenResponse, Error> {
    let client = Client::new(client_id, client_secret, auth_url, token_url);
    let (csrf_token, pkce_code_verifier) = user_db.get_user_oauth_state(user);
    if csrf_token != state {
        return Err(...);
    }

    let params =
        ExchangeParams::new(code)
            .withPkce(pkce_code_verifier);
    let extensions = ...;
    return client.exchange_code_extension(params, extensions);
}
```
(similarly, the `authorize_url` and `exchange_code` functions would accept their respective `*Params` types but without the additional extensions)

With this approach, we can add new extensions into this library without breaking backward compatibility by adding new methods to `*Params` and `AuthorizationRequest`. Clients that want to build a higher-level "session" abstraction on top of this can still do so, but other clients can take advantage of common extensions without having to build a session abstraction or manually pass in request parameters to `authorize_url_extension` and `exchange_code_extension`.

Thoughts?

jongiddy:
OK, I think there's a nice abstraction there somewhere, but it looks like it would add a lot into this library, and might be tricky to get right first time. I think a few tweaks to the `CodeVerifier` object might give a more natural lower-level approach, and allow me to remove the PKCE specific methods.

jongiddy:
It's take a while for me to get back to this, but I've rewritten the PKCE code to only use `_extension` methods, making it more composable with other extensions.

ramosbugs:
Thanks again!

Issue #44, opened 2018-10-07
Async API


thedrow:
Currently this library uses curl to perform HTTP requests.
It'd be nice if we could add an async API that supports the tokio event loop.
I'm not sure if it's possible with the curl bindings but it will allow us to use the async/await syntax in the future.

ramosbugs:
Thanks for the suggestion! I would definitely be interested in adding async support to this library, but I probably won't have the bandwidth to implement these changes any time soon. I would happily review any contributions toward this effort, however.

As far as continuing to use curl, that's not a strict requirement if there are more tokio-friendly HTTP clients that still provide reasonable security guarantees (e.g., TLS cert validation). I haven't looked into other Rust HTTP clients at all.

thedrow:
Reqwest (https://docs.rs/reqwest/0.9.2/reqwest/) seems pretty popular. It also has an async API.

ramosbugs:
I continue to agree with the goals of supporting an async API and different (non-curl) HTTP backends. However, it doesn't look like we're close to merging those features in, so I'm going to go ahead and promote the current master branch to 2.0.0-beta.1 and avoid any additional breaking changes on the 2.x major release.

We can still merge these new features in as soon as they're ready in preparation for a breaking 3.x release.

udoprog:
I have an async branch of this library here (backed by reqwest):
https://github.com/udoprog/oauth2-rs/tree/async

I'm using it in one of my projects here:
https://github.com/udoprog/setmod/blob/master/bot/src/oauth2.rs#L420

The biggest difference is that setting up a request returns a request builder to reduce the amount of parameterization needed. Instead, you specify the expected response type during a separate `execute()` function which returns a future:
https://github.com/udoprog/setmod/blob/master/bot/src/oauth2.rs#L139

My project includes examples for interacting with Twitch, YouTube, and Spotify. Each of with their own quirks.

If you're interested in moving towards something like this in `3.x`, I would be happy to clean up the patches and send a PR.

ramosbugs:
Thanks, I'll take a look in the next week or so!

ramosbugs:
I'm working on porting over @udoprog's changes but with some alterations, such as:
- not relying on nightly features like async await (this crate needs to work on stable, though I'm not opposed to some nightly-only features gated by feature flags)
- abstracting away the HTTP implementation so that other backends can be plugged in (I'll probably make Reqwest the new default, and include curl as an alternative behind a feature flag)
[...]

PR #46, opened 2018-10-26
async API

legokichi:
https://github.com/ramosbugs/oauth2-rs/issues/44#issuecomment-427739934

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -225,6 +225,7 @@ use std::marker::{Send, Sync, PhantomData};
 use std::ops::Deref;
 use std::time::Duration;
 
+use url::percent_encoding::{utf8_percent_encode, EncodeSet};
 use futures::prelude::*;
 use failure::{Backtrace, Fail};
 use rand::{thread_rng, Rng};
@@ -235,6 +236,24 @@ use url::Url;
 use prelude::*;
 
 const CONTENT_TYPE_JSON: &str = "application/json";
+
+#[derive(Clone)]
+/// This struct is used to maintain the strict URI encoding standard as proposed by RFC 3986
+struct StrictEncodeSet;
+
+impl EncodeSet for StrictEncodeSet {
+    #[inline]
+    fn contains(&self, byte: u8) -> bool {
+        let upper = byte >= 0x41 && byte <= 0x5a;
+        let lower = byte >= 0x61 && byte <= 0x7a;
+        let numeric = byte >= 0x30 && byte <= 0x39;
+        let hyphen = byte == 0x2d;
+        let underscore = byte == 0x5f;
+        let tilde = byte == 0x7e;
+        let period = byte == 0x2e;
+        !(upper || lower || numeric || hyphen || underscore || tilde || period)
+    }
+}
 
 ///
 /// Indicates whether requests to the authorization server should use basic authentication or
@@ -897,10 +916,12 @@ impl<EF: ExtraTokenFields + Send + 'static, TT: TokenType + Send + 'static, TE: 
                 // Section 2.3.1 of RFC 6749 requires separately url-encoding the id and secret
                 // before using them as HTTP Basic auth username and password. Note that this is
                 // not standard for ordinary Basic auth, so curl won't do it for us.
-                let username = self.client_id.as_str();
+                let username = utf8_percent_encode(self.client_id.as_str(), StrictEncodeSet);
                 let password = match &self.client_secret {
                     None => None,
-                    Some(ref o) => Some(o.secret().as_str()),
+                    Some(ref o) => Some(
+                        utf8_percent_encode(o.secret().as_str(), StrictEncodeSet)
+                    ),
                 };
                 req_builder = req_builder.basic_auth(username, password);
             }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -621,6 +621,7 @@ pub struct Client<EF: ExtraTokenFields, TT: TokenType, TE: ErrorResponseType> {
     token_url: Option<TokenUrl>,
     scopes: Vec<Scope>,
     redirect_url: Option<RedirectUrl>,
+    http_client: ::reqwest::async::Client,
     phantom_ef: PhantomData<EF>,
     phantom_tt: PhantomData<TT>,
     phantom_te: PhantomData<TE>,
@@ -661,6 +662,7 @@ impl<EF: ExtraTokenFields + Send + 'static, TT: TokenType + Send + 'static, TE: 
             token_url,
             scopes: Vec::new(),
             redirect_url: None,
+            http_client: ::reqwest::async::Client::new(),
             phantom_ef: PhantomData,
             phantom_tt: PhantomData,
             phantom_te: PhantomData,
@@ -902,7 +904,7 @@ impl<EF: ExtraTokenFields + Send + 'static, TT: TokenType + Send + 'static, TE: 
         token_url: &TokenUrl,
         mut params: Vec<(&'b str, &'a str)>
     ) -> impl Future<Item = RequestTokenResponse, Error = reqwest::Error> + Send + 'static {
-        let mut req_builder = ::reqwest::async::Client::new().post(&token_url.to_string()[..]);
+        let mut req_builder = self.http_client.post(&token_url.to_string()[..]);
 
         // FIXME: add support for auth extensions? e.g., client_secret_jwt and private_key_jwt
         match self.auth_type {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -212,12 +212,13 @@
 
 extern crate base64;
 extern crate futures;
-extern crate reqwest;
 extern crate failure;
 extern crate rand;
+extern crate reqwest;
 extern crate serde;
 #[macro_use] extern crate serde_derive;
 extern crate serde_json;
+extern crate tokio;
 extern crate url;
 
 use std::fmt::{Debug, Display, Error as FormatterError, Formatter};
@@ -225,12 +226,12 @@ use std::marker::{Send, Sync, PhantomData};
 use std::ops::Deref;
 use std::time::Duration;
 
-use url::percent_encoding::{utf8_percent_encode, EncodeSet};
 use futures::prelude::*;
 use failure::{Backtrace, Fail};
 use rand::{thread_rng, Rng};
 use serde::Serialize;
 use serde::de::DeserializeOwned;
+use url::percent_encoding::{utf8_percent_encode, EncodeSet};
 use url::Url;
 
 use prelude::*;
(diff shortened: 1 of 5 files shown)


tests/lib.rs
```diff
@@ -273,15 +276,18 @@ fn test_exchange_code_successful_with_complete_json_response() {
 fn test_exchange_client_credentials_with_basic_auth() {
     let mock = mock("POST", "/token")
         // base64(urlencode("aaa/;&") + ":" + urlencode("bbb/;&"))
-        .match_header("Authorization", "Basic YWFhJTJGJTNCJTI2OmJiYiUyRiUzQiUyNg==")
+        .match_header("Authorization", "basic YWFhJTJGJTNCJTI2OmJiYiUyRiUzQiUyNg==")
```
legokichi: reqwest puts lowercase letters. 
https://docs.rs/reqwest/0.9.3/src/reqwest/async_impl/request.rs.html#171

legokichi:
reqwest does not have HttpConnector API - 
 https://docs.rs/hyper/0.12.12/hyper/client/struct.HttpConnector.html

ramosbugs:
thanks for the PR! I'm traveling this weekend, but I'll take a look next week.

src/lib.rs
```diff
@@ -236,6 +237,24 @@ use prelude::*;
 
 const CONTENT_TYPE_JSON: &str = "application/json";
 
+#[derive(Clone)]
+/// This struct is used to maintain the strict URI encoding standard as proposed by RFC 3986
+struct StrictEncodeSet;
```
legokichi: https://github.com/rusoto/rusoto/blob/7640ca3de1e6486b0b25bce395810a9434a66ee5/rusoto/core/src/signature.rs#L593

ramosbugs:
Sorry for the delay! Looking this PR over now.

> I'm thinking of using hyper instead of reqwest.
> reqwest does not have HttpConnector API -
> https://docs.rs/hyper/0.12.12/hyper/client/struct.HttpConnector.html

I think it's definitely useful (often required) for clients to be able to specify connection parameters, including TLS settings (trusted CA certs, cipher suites, etc.) and proxy configs. This functionality has been missing from this crate so far, so thank you for bringing it up!

However, it doesn't look like `hyper::client::HttpConnector` really lets us specify these parameters. Rather, that interface seems to deal with lower level network options. What sort of customization were you hoping to support via `HttpConnector`?

As an alternative, `reqwest::async::ClientBuilder` does seem to allow significant customization: https://docs.rs/reqwest/0.9.4/reqwest/async/struct.ClientBuilder.html. What if we add a `set_http_client` method to `oauth2::Client` so that users can pass in their own `reqwest::async::Client`s?

> - https://actix.rs/actix-web/actix_web/client/struct.ClientConnector.html#method.with_connector

it looks like this one refers to a different interface (`openssl`'s `SslConnector` struct rather than `hyper`'s `Connect` trait)

---

I'll submit comments on the rest of the PR soon.

src/lib.rs
ramosbugs: I'm finding the chain of specs referenced by https://tools.ietf.org/html/rfc6749#appendix-B to be a bit hard to follow, so it's not super clear to me exactly which characters need to be escaped (especially since we're dealing with an HTTP header here, rather than a URI). This code does seem to maintain the current behavior as implemented in https://github.com/curl/curl/blob/master/lib/escape.c#L40, but I wonder if it would suffice to use  
[`USERINFO_ENCODE_SET`](https://docs.rs/percent-encoding/1.0.1/percent_encoding/struct.USERINFO_ENCODE_SET.html), which seems to be intended for username and password escaping (according to https://github.com/dtolnay/url/blob/master/UPGRADING.md).

@DeeUnderscore do you have any thoughts here, as the author of #41? (see usage below)

tests/lib.rs
```diff
@@ -202,8 +203,9 @@ fn test_exchange_code_successful_with_minimal_json_response() {
             Url::parse(&(SERVER_URL.to_string() + "/token")).unwrap(),
         )),
     );
-    let token = client
-        .exchange_code(AuthorizationCode::new("ccc".to_string()))
+    let token = ::tokio::runtime::Runtime::new().unwrap().block_on(
```
ramosbugs: since each test needs to do this, let's define a function that takes a future, creates a runtime, blocks on it, and returns the result. it's probably worth trying to minimize the test boilerplate.

src/lib.rs
DeeUnderscore: By my reading, this adheres to `application/x-www-form-urlencoded`—everything outside of the permitted bytes is encoded. [This WHATWG spec](https://url.spec.whatwg.org/#application/x-www-form-urlencoded) seems to be essentially equivalent to what RFC6749 wanted when it described the encoding in Appendix B. 

URLs can use forms of percent encoding which aren't `application/x-www-form-urlencoded`, since they permit more byte ranges (presumably to make URLs more readable). `USERINFO_ENCODE_SET` ([WHATWG ref](https://url.spec.whatwg.org/#userinfo-percent-encode-set)), in particular, is intended for the userinfo component of a URL, ie. the `user:password` in `https://user:password@example.org/`.

In practical terms, you could probably get away with a less strict encoding, and I suspect most OAuth implementations in the wild avoid using characters which could cause problems anyway. But, for closest adherence to the standard I'd recommend what this PR is doing.

ramosbugs: Thanks for the reference and rationale! That makes sense; I agree with the recommendation.

The encoding here does seem to deviate ever so slightly from the WHATWG spec, though:
- spaces encoded with `%20` instead of `+`
- asterisks (`0x2A`) excluded from the encoding set but mentioned by WHATWG
- tildes (`0x7E`) included in encoding set but excluded by WHATWG

None of these deviations seems particularly problematic, though it's probably safer for interoperability to err on the side of percent-encoding more values rather than fewer. Unless there's an error in the WHATWG spec, we should probably remove tildes from this set.

src/lib.rs
```diff
@@ -807,7 +828,7 @@ impl<EF: ExtraTokenFields, TT: TokenType, TE: ErrorResponseType> Client<EF, TT,
     pub fn exchange_code(
         &self,
         code: AuthorizationCode
-    ) -> Result<TokenResponse<EF, TT>, RequestTokenError<TE>> {
+    ) -> impl Future<Item = TokenResponse<EF, TT>, Error = RequestTokenError<TE>> + Send + 'static {
```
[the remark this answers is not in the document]
legokichi: tokio requires [`send + 'static`](https://tokio.rs/blog/2018-03-tokio-runtime/) because it adopts the [work-stealing algorithm](https://tokio-rs.github.io/tokio/tokio/runtime/fn.run.html).

src/lib.rs
```diff
@@ -808,6 +841,14 @@ impl<EF: ExtraTokenFields, TT: TokenType, TE: ErrorResponseType> Client<EF, TT,
         &self,
         code: AuthorizationCode
     ) -> Result<TokenResponse<EF, TT>, RequestTokenError<TE>> {
+        block_on(self.exchange_code_async(code))
+    }
+
+    /// async edition
+    pub fn exchange_code_async(
```
legokichi: futures trait also [changes greatly](https://rust-lang-nursery.github.io/futures-rs/) in [Rust 2018](https://rust-lang-nursery.github.io/edition-guide/rust-2018/control-flow/async-await-for-easier-concurrency.html).
How about doing this as a provisional response for the past few months?

* exchange_code_async
* exchange_password_async
* exchange_client_credentials_async
* exchange_refresh_token_async

legokichi:
Here are the implementations of https_connector.

* native-tls: https://github.com/hyperium/hyper-tls
* rustls: https://github.com/ctz/hyper-rustls

src/lib.rs
legokichi: Alternative choice: just constructs and returns [`http::request::Request`](https://docs.rs/http/0.1.13/http/request/struct.Request.html) object.
User can chose own http request library e.g. curl, reqwest, hyper, actix-web ...etc.

src/lib.rs
```diff
@@ -211,31 +211,62 @@
...
+    #[inline]
+    fn contains(&self, byte: u8) -> bool {
+        let upper = byte >= 0x41 && byte <= 0x5a;
+        let lower = byte >= 0x61 && byte <= 0x7a;
+        let numeric = byte >= 0x30 && byte <= 0x39;
+        let hyphen = byte == 0x2d;
+        let underscore = byte == 0x5f;
+        let tilde = byte == 0x7e;
+        let period = byte == 0x2e;
+        !(upper || lower || numeric || hyphen || underscore || tilde || period)
```
ramosbugs: based on the conversation above, I think we should probably remove `tilde`

src/lib.rs
ramosbugs: I think it's fine to put the pre-Rust 2018 async functionality into this crate. However, I think we should do this via a separate `Client` implementation (i.e., `oauth2::async::Client`). Then, when we're ready to add support for Rust 2018 async, we can call that `oauth2::async2018::Client` or something. Each of these can present the same `exchange_*` functions, just with different return types. This change will probably require some refactoring to share as much code as possible between the different `Client` implementations.

ramosbugs:
> Here are the implementations of https_connector.
> 
> * native-tls: https://github.com/hyperium/hyper-tls
> * rustls: https://github.com/ctz/hyper-rustls
> * https://github.com/pimeys/hyper-alpn
> * https://github.com/sfackler/hyper-native-tls
> * https://github.com/sfackler/hyper-openssl

Ohh, I see. So it looks like what we really want to accept is a struct that implements the `hyper::client::connect::Connect` trait, which can do anything it wants as long as it yields a `(hyper::client::connect::Connect::Transport, hyper::client::connect::Connected)`.  Something like what you mentioned above should work, e.g.:
```
pub struct Client<C: Connect, EF: ExtraTokenFields, TT: TokenType, TE: ErrorResponseType> {
    connector: C,
    ...
}

// new() sets a sensible default (and secure) connector (require HTTPS, etc.)

pub fn with_connector(mut self, connector: C) -> Self {
    self.connector = connector;

    self
}
```

bachp:
I would like to use this in the context of actix-web. So I would be very happy if there is a way to specify the client library to use (reqwest, actix, ...). Did you already give any toughts on how this could be done?

If it would be desired to have this as a feature I might be able to come up with a PR that allows to select actix-web as a replacement for reqwest.

ramosbugs:
> I would like to use this in the context of actix-web. So I would be very happy if there is a way to specify the client library to use (reqwest, actix, ...). Did you already give any toughts on how this could be done?

I think the simplest approach to supporting multiple HTTP clients is probably via a trait. Since the `hyper::client::connect::Connect` trait is quite low level, I think it's a good candidate. This also lets people use `hyper` by default, which seems to have a significantly larger volume of downloads on crates.io than `reqwest` or `actix-web`. Other client libraries could potentially be used by writing wrappers that implement the `hyper::client::connect::Connect` trait on top of those libraries.

Alternatively, we could probably do something at compile time based on feature flags, but I haven't looked into that much, and it feels like that would be harder to maintain over time.

Was there specific functionality from `actix-web` that you were looking to use in conjunction with this crate? Understanding the use case a bit more might help illuminate the right solution.

I think it's also probably worth filing a separate Github issue for this so that we can refocus #44 on the async discussion, and the new issue on the need for a generic HTTP client interface.

ramosbugs:
latest master now features an async API with pluggable HTTP clients

PR #47, opened 2018-12-03
Add optional scope to password grant

marcelbuesing:
This adds the optional scope parameter as stated in https://tools.ietf.org/html/rfc6749#section-4.3.2


ramosbugs:
Thanks for the PR!

PR #53, opened 2019-02-28
Bump mockito to 0.16.0 to have a working Windows build

udoprog:
Windows support was fixed in a later version of mockito. Bumping it makes the tests pass.


ramosbugs:
Thanks!

Issue #59, opened 2019-05-09
Can't override error type for client for servers that don't adhere to spec.


denniscollective:
The request token Error wraps the ```ErrorResponseType``` in ```ErrorResponse```

```rust
pub enum RequestTokenError<T: ErrorResponseType + Send + Sync + 'static> {
    ServerResponse(ErrorResponse<T>),
   ...
}
```

I'm integrating with an external oauth provider that returns errors with the following shape
```rust
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
 pub struct CustomErrorResponse {
      message: String,
      errors: Vec<serde_json::Value>,
}
```
I tried making a client with my ```CustomErrorResponse``` as the ```ErrorResponseType```

```rust
type OauthClient = Client<CustomErrorResponse, CustomTokenResponse, BasicTokenType>;
impl oauth2::ErrorResponseType for models::ErrorResponse {}
```

but since the ```RequestTokenError::ServerResponse``` wraps it in an ```ErrorResponse``` it bubbles up into a ```RequestTokenError::Parse```.

Could the wrapping be moved up out of the enum to allow for servers that return their own format for errors?

Is there another affordance for this in the library somewhere that I'm missing?

Thanks!

ramosbugs:
Unfortunately, arbitrary error responses aren't supported by the library. The spec is pretty clear about the error response format: https://tools.ietf.org/html/rfc6749#section-5.2. The extensibility in the library allows for additional error codes (as mentioned in https://tools.ietf.org/html/rfc6749#section-11.4), but not an entirely different serialization format.

If you're not in a position to be able to change the server error response format, my recommendation is to manually parse the contents returned in `RequestTokenError::Parse`, which should let you read the error response in an arbitrary format. One enhancement to facilitate this would be to include the HTTP response code in `RequestTokenError::Parse`, but that would be a breaking change that would need to wait for a 3.x version bump.

denniscollective:
I can understand hesitancy to change things if it would require a 3.x bump, I thought things could still be changed because 2.0 hasn't actually been released yet.

It is a bummer that interfacing with a non-standards-compliant external provider is so complicated, but I'm powerless to change things on there side. There is an asymmetry in that the library arbitrary control over parsing of successful responses, but does not allow the same control for error handling.

[...]

PR #60, opened 2019-05-12
Client can parse non-standard Error Responses

denniscollective:
ErrorResponse is now the trait that client takes.
The ErrorResponse struct has been renamed to StandardErrorResponse.
StandardErrorResponse's parameterization of ErroResponseType is it's own concern.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -237,7 +237,7 @@ use std::time::Duration;
 use curl::easy::Easy;
 use rand::{thread_rng, Rng};
 use serde::de::DeserializeOwned;
-use serde::{Serialize};
+use serde::Serialize;
 use sha2::{Digest, Sha256};
 use url::Url;
 
@@ -1418,9 +1418,11 @@ where
 ///
 /// Server Error Response
 ///
-/// See `StandardErrorResponse` for an implementation adhering to the oauth standard.
+/// This trait exists separately from the `StandardErrorResponse` struct
+/// to support customization by clients, such as supporting interoperability with
+/// non-standards-complaint OAuth2 providers
 ///
-pub trait ErrorResponse: DeserializeOwned + Serialize + Send + Sync + Debug + Display + 'static {}
+pub trait ErrorResponse: Debug + DeserializeOwned + Display + Send + Serialize + Sync {}
 
 ///
 /// Error types enum.
@@ -1529,7 +1531,7 @@ impl<TE: ErrorResponseType> Display for StandardErrorResponse<TE> {
 /// Error encountered while requesting access token.
 ///
 #[derive(Debug, Fail)]
-pub enum RequestTokenError<T: ErrorResponse> {
+pub enum RequestTokenError<T: ErrorResponse + 'static> {
     ///
     /// Error response returned by authorization server. Contains the parsed `ErrorResponse`
     /// returned by the server.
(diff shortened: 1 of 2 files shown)


denniscollective:
Fixes #59.

I'm not exactly sure about the definition of the ErrorResponseTrait. Specifically I wasn't sure about Deserialize vs DeserializeOwned and also Send + Sync + Debug + Display vs Fail.

I renamed the ErrorResponse struct to mirror StandardErrorResponse. I'm down to rename these things to anything that you think would be easiest.

ramosbugs:
thanks for the PR! still traveling through this weekend, but I'll take a close look next week.

src/lib.rs
```diff
@@ -237,7 +237,7 @@ use std::time::Duration;
 use curl::easy::Easy;
 use rand::{thread_rng, Rng};
 use serde::de::DeserializeOwned;
-use serde::Serialize;
+use serde::{Serialize};
```
ramosbugs: nit: let's remove the curly braces

actually, mind running `cargo +nightly fmt`? there are a few minor formatting issues that I think will be fixed by that command

src/lib.rs
```diff
@@ -1415,6 +1415,13 @@ where
     }
 }
 
+///
+/// Server Error Response
+///
+/// See `StandardErrorResponse` for an implementation adhering to the oauth standard.
```
ramosbugs: nit: OAuth

I think it's worth mentioning here that the trait exists to support interoperability with non-compliant providers. something similar to the rustdoc for `TokenResponse`.

src/lib.rs
```diff
@@ -1415,6 +1415,13 @@ where
...
+///
+pub trait ErrorResponse: DeserializeOwned + Serialize + Send + Sync + Debug + Display + 'static {}
```
ramosbugs: I think we can remove `Display` here as a trait bound. `StandardErrorResponse` implements it, but it's not a strict requirement that all error response types also do that. the other bounds seem reasonable, though.

nit: let's sort these alphabetically

src/lib.rs
```diff
@@ -1436,7 +1443,7 @@ pub trait ErrorResponseType:
 /// authentication schemes and extensions.
 ///
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
-pub struct ErrorResponse<T: ErrorResponseType> {
+pub struct StandardErrorResponse<T: ErrorResponseType> {
```
ramosbugs: how about `+ 'static` here so we can remove it from the impl below?

tests/lib.rs
```diff
@@ -1142,6 +1142,73 @@ fn test_extension_with_simple_json_error() {
     );
 }
 
+mod custom_errors {
```
ramosbugs: nice test!

src/lib.rs
```diff
@@ -1520,13 +1529,13 @@ impl<TE: ErrorResponseType> Display for ErrorResponse<TE> {
...
 #[derive(Debug, Fail)]
-pub enum RequestTokenError<T: ErrorResponseType + Send + Sync + 'static> {
+pub enum RequestTokenError<T: ErrorResponse> {
     ///
     /// Error response returned by authorization server. Contains the parsed `ErrorResponse`
     /// returned by the server.
     ///
     #[fail(display = "Server returned error response `{}`", _0)]
-    ServerResponse(ErrorResponse<T>),
+    ServerResponse(T),
```
ramosbugs: oh I guess this requires `Display`

denniscollective:
I believe the most recent commit should have fixed your formatting concerns.

- I moved the 'static bounds to RequestTokenError definition because that was where it was actually needed (not for StandardErrorResponse)

- The same could not be done for the Display bounds because there were other places it was needed

Issue #63, opened 2019-06-11
Support non compliant providers


sennav:
Hi,

I am writing a CLI for wunderlist.com and using this library to do the oauth authentication. It was everything going well until I realised that wunderlist is not fully compliant with the oauth specs. In their `access_token` response they do not set `token_type` and thus oauth2 fails to transform their response into an object with a parsing error.

What do you think about making `token_type` optional or allowing users to make it optional. I'm a beginner in rust and to solve my issue I simply commented out the `token_type` and I'm using oauth2 that I compile locally but I assume this will happen with other services. I would be happy to help with a PR if you can give me some guidance on how and if you ever want to support this.

ramosbugs:
Hi @sennav,

Unfortunately `token_type` is a mandatory response parameter according to https://tools.ietf.org/html/rfc6749#section-5.1, so I don't think it would be appropriate to make it optional in this library (forcing users to handle `Option` in places that shouldn't be `None`).

That said, the [`TokenResponse`](https://docs.rs/oauth2/2.0.0-beta.2/oauth2/trait.TokenResponse.html) trait allows you to implement your own token response type instead of using the `StandardTokenResponse` struct that is included in the crate. You'll still need the `token_type` method to return something, which could either be a default value you choose or you could try making the `TT` type parameter be an `Option` (not sure if that will work).

Hope that helps!

sennav:
Hi, yes you are right, it would not make sense to force all users to handle an `Option` because some providers are not compliant. I did not know I had the option that you suggested, so thanks for that.

With that said, I'm trying to implement my own token response as you mentioned and I think I got pretty far, but I'm still struggling a bit with rust (as I said, I'm a beginner). Would it be ok to try to submit an example of that then? I guess more beginners like me will try to do something similar.

ramosbugs:
sure, but I probably won't have much time to spend giving general Rust help

PR #65, opened 2019-07-02
Add an example of a non compliant oauth provider (wunderlist)

sennav:
Continuing the suggestion from #63 

Wunderlist does not return the token type as specified in RFC 6749
section 7.1 https://tools.ietf.org/html/rfc6749#section-7.1 so the
wunderlist example showcases how one can create new types to allow for
those situations.

Let me know if there's anything you wish to improve.

### DIFF
--- a/examples/wunderlist.rs
+++ b/examples/wunderlist.rs
@@ -1,0 +1,248 @@
+//!
+//! This example showcases the Wunderlist OAuth2 process for requesting access to the user's todo lists.
+//! Wunderlist does not implement the correct token response, so this serves as an example of how to
+//! implement a custom client.
+//!
+//! Before running it, you'll need to create your own wunderlist app.
+//!
+//! In order to run the example call:
+//!
+//! ```sh
+//! WUNDER_CLIENT_ID=xxx WUNDER_CLIENT_SECRET=yyy cargo run --example wunderlist
+//! ```
+//!
+//! ...and follow the instructions.
+//!
+
+extern crate base64;
+extern crate oauth2;
+extern crate rand;
+extern crate url;
+extern crate serde;
+
+use oauth2::TokenType;
+use oauth2::basic::{ BasicErrorResponse, BasicTokenType };
+use oauth2::curl::http_client;
+use oauth2::helpers;
+use oauth2::{
+    AccessToken,
+    AuthorizationCode,
+    AuthUrl,
+    Client,
+    ClientId,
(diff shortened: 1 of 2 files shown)


examples/wunderlist.rs
```diff
@@ -0,0 +1,258 @@
...
+}
+
+impl<EF> NonStandardTokenResponse<EF>
+where
+    EF: ExtraTokenFields,
+{
+    ///
+    /// Get token
+    ///
+    pub fn get_access_token(self) -> AccessToken {
```
ramosbugs: it doesn't look like we need this function, since users can just call `.access_token().clone()` if they need an owned version, right?

examples/wunderlist.rs
```diff
@@ -0,0 +1,258 @@
...
+/// [Section 5.1 of RFC 6749](https://tools.ietf.org/html/rfc6749#section-5.1), as well as
+/// extensions defined by the `EF` type parameter.
+/// In this particular example token_type is optional to showcase how to deal with a non
+/// compliant provider.
+///
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+pub struct NonStandardTokenResponse<EF: ExtraTokenFields> {
+    access_token: AccessToken,
+    #[serde(default = "default_token_type")]
+    token_type: Option<BasicTokenType>,
```
ramosbugs: it's probably worth having a comment here calling out that this is the part where Wunderlist deviates from the spec

ramosbugs:
oh, be sure to address the Travis CI failures too

examples/wunderlist.rs
sennav: Done, thanks for pointing it out.

examples/wunderlist.rs
sennav: Added the comment please have a look.

sennav:
It seems the travis build failed on nightly because it hanged for more than 10 min, I'm not sure what this could be but I think it's unrelated with the PR. Can you try to rerun it? Thanks

ramosbugs:
It was complaining about `PartialEq` before: https://travis-ci.org/ramosbugs/oauth2-rs/jobs/553459777. I've been seeing some CI failures on nightly due to some issue with the cargo caching. I just deleted the caches again off of Travis. I'll go ahead and merge and hope the next run passes.

sennav:
Thanks!

PR #68, opened 2019-09-13
Make curl an optional dependency

mwilliammyers:
Make reqwest the default http backend.

Discussed in [#44](https://github.com/ramosbugs/oauth2-rs/issues/44#issuecomment-501591680).


ramosbugs:
thanks for the PR!

PR #73, opened 2019-11-22
Add Letterboxd API example with custom request signing.

boxdot:
Letterboxd: https://letterboxd.com/
API docs: http://api-docs.letterboxd.com/


ramosbugs:
Thanks!

PR #74, opened 2019-11-27
add features to select TLS alternatives

KeenS:
I'd like to select TLS library.


ramosbugs:
thanks for the PR! however, I think clients of the library who want a different TLS version should just disable the `reqwest` feature flag and instead provide their own HTTP client that adheres to `FnOnce(HttpRequest) -> Result<HttpResponse, RE>`.

the `reqwest` module in this library is really just a convenience; I'd rather not go down the path of supporting all the different combinations of HTTP client feature flags.

KeenS:
> the reqwest module in this library is really just a convenience

I see. It's reasonable. Then I close this P-R. Thank you for your reply.

Issue #79, opened 2019-12-18
reqwest 0.10 and http 0.2


dthul:
reqwest 0.10 just switched to http 0.2 (unreleased as of yet, reqwest 0.10.0-alpha.2 still uses http 0.1).

I saw that a lot of feature refactoring just happened in this project to support all combinations of futures 0.1 / 0.3 and reqwest 0.9 / 0.10. I don't know what's the best option to handle the version of the `http` crate though. I see several options:

* The version of `http` could either be bound to the selected reqwest version (0.2 for reqwest 0.10 and 0.1 for reqwest 0.9), or
* There could be another feature flag pair `http_0_1` and `http_0_2`, or
* This crate just switches to http 0.2 altogether

I feel that the last option might be the best. It would require some `http 0.1` <-> `http 0.2` type translation code in case reqwest 0.9 is used.

Anyway, I just wanted to open this issue to point out this upcoming incompatibility.

ramosbugs:
if possible, I'd like to switch entirely to `http` 0.2, as long as that works with both `reqwest` versions. I would have preferred to switch to `reqwest` 0.10 entirely as well, but it doesn't maintain backward compatibility with `futures` 0.1.

ramosbugs:
I tried upgrading to `reqwest` 0.10 and `http` 0.2 in 81e2a87de76b351ab97fb74548d1e3cbe86cd923, but that breaks compatibility with rustc < 1.39 because `http` 0.2 depends on `bytes` 0.5, which uses newer language features not supported by older compiler versions.

for now, I think I'm going to leave the external HTTP client interface as `http` 0.1, and have the `reqwest` 0.10 client convert to `http` 0.2 types internally

ramosbugs:
e.g.:
```
error: `core::slice::<impl [T]>::len` is not yet stable as a const fn
   --> /Users/daramos/.cargo/registry/src/github.com-1ecc6299db9ec823/bytes-0.5.3/src/bytes.rs:121:18
    |
121 |             len: bytes.len(),
    |                  ^^^^^^^^^^^
```

dthul:
Seems like a reasonable approach. I have some crude version of that `http` type conversion code in my own project (since I already use `reqwest` 0.10 with this crate through a custom async client implementation). I unfortunately won't have time in January to help you out with that though.

gzp-crey:
Just from curiosity. As it is a major version change, why not update the minimal required rust version as well ?
[...]

PR #82, opened 2020-01-20
bind reqwest 0.9 to http 0.1 and reqwest 0.10 to http 0.2

gzp-crey:
This is an alternative solution for #79 :

- reqwest-09 requires http 0.1 (and http 0.1 is exported as http)
- reqwest-010 requires http 0.2 (and http 0.2 is exported as http)

Also the default feature was turned off, as 
 - it would require reqwest-010 feature would require a no-default
 - request is "just" a sample client implementation and on could provide his/her implementation if required.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,11 +15,11 @@ all-features = true
 maintenance = { status = "actively-developed" }
 
 [features]
-default = ["reqwest-09"]
+default = []
 futures-01 = ["futures-0-1"]
 futures-03 = ["futures-0-3", "async-trait"]
-reqwest-09 = ["reqwest-0-9"]
-reqwest-010 = ["reqwest-0-10"]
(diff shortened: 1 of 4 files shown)


ramosbugs:
thanks for the PR! however, I don't think it makes sense to tie the `http` version to the `reqwest` version because the `http` interface needs to be exposed (via `HttpRequest` and `HttpResponse` for custom HTTP clients) even if neither `reqwest` version is enabled.

I also want to keep `reqwest 0.9` enabled by default since this crate should work "out of the box" for most users without requiring them to add any specific feature flags. It can easily be disabled for those who want a different HTTP client by disabling the default features, as outlined in https://docs.rs/openidconnect/1.0.0-alpha.11/openidconnect/#importing-openidconnect-selecting-an-http-client-interface.

PR #83, opened 2020-03-04
Make http 0.1 optional

kaj:
Hi @ramosbugs , thanks for your work on oauth2 and openidconnect in rust!

Here's a suggestion for some dependency cleanup (related to ramosbugs/openidconnect-rs#8):

Make the http 0.1 dependency optional and enable it for reqwest 0.9. Reexport the enabled http version as oauth::http (rather than exporting http 0.1 even if using reqwest 0.10, which uses http 0.2).

Addendum:  I just found #82, which is rather similar to this, but perhaps the timing is better now that reqwest 0.10 is actually released?

Or maybe support for reqwest 0.9, futures 0.1 and http 0.1 should just be dropped before oauth2 3.0 is actually released?  That would simplify things quite a bit, but it may be a bit premature?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ maintenance = { status = "actively-developed" }
 
 [features]
 default = ["reqwest-09"]
-futures-01 = ["futures-0-1"]
+futures-01 = ["futures-0-1", "http-0-1"]
 futures-03 = ["futures-0-3", "async-trait"]
 reqwest-09 = ["reqwest-0-9", "http-0-1"]
 reqwest-010 = ["reqwest-0-10", "http-0-2"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ maintenance = { status = "actively-developed" }
 default = ["reqwest-09"]
 futures-01 = ["futures-0-1"]
 futures-03 = ["futures-0-3", "async-trait"]
-reqwest-09 = ["reqwest-0-9"]
+reqwest-09 = ["reqwest-0-9", "http-0-1"]
 reqwest-010 = ["reqwest-0-10", "http-0-2"]
 
 [dependencies]
@@ -29,7 +29,7 @@ failure = "0.1"
 failure_derive = "0.1"
 futures-0-1 = { version = "0.1", optional = true, package = "futures" }
 futures-0-3 = { version = "0.3", optional = true, package = "futures" }
-http = "0.1"
+http-0-1 = { version = "0.1", optional = true, package = "http" }
 http-0-2 = { version = "0.2", optional = true, package = "http" }
 rand = "0.6"
 reqwest-0-9 = { version = "0.9", optional = true, package = "reqwest" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ maintenance = { status = "actively-developed" }
 [features]
 default = ["reqwest-09"]
 futures-01 = ["futures-0-1", "http-0-1"]
-futures-03 = ["futures-0-3", "async-trait"]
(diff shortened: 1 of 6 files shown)


ramosbugs:
thanks for the PR! if we merge this change, it doesn't look like there's a way for users who want to define their own blocking HTTP client to use this crate without enabling either `request-0-9` or one of the `futures` features. I think we need to continue supporting that use case without enabling unnecessary dependencies.

> Or maybe support for reqwest 0.9, futures 0.1 and http 0.1 should just be dropped before oauth2 3.0 is actually released? That would simplify things quite a bit, but it may be a bit premature?

My goal for 3.0 has been to continue supporting rustc back to 1.34, which can't compile http 0.2 (or futures 0.3). That goal may be misguided, though.

I think dropping support for futures 0.1 would be a bigger problem than dropping support for rustc 1.34, though. 81e2a87de76b351ab97fb74548d1e3cbe86cd923 was my attempt to migrate to http 0.2 while still supporting both `requests` and `futures` versions. I reverted it after realizing it broke 1.34.

Do you know if there's any data on which rustc versions are still in use? I think maybe restoring and fixing up 81e2a87de76b351ab97fb74548d1e3cbe86cd923 may be the best way to proceed given the amount of interest in upgrading to http 0.2.

kaj:
> if we merge this change, it doesn't look like there's a way for users who want to define their own blocking HTTP client to use this crate without enabling either `request-0-9` or one of the `futures` features. I think we need to continue supporting that use case without enabling unnecessary dependencies.

Can't they in that case enable`http-0-1` or `http-0-2` directly as a feature?

> My goal for 3.0 has been to continue supporting rustc back to 1.34, which can't compile http 0.2 (or futures 0.3). That goal may be misguided, though.

How about dropping support fore futures 0.1 and keeping synchronous-only support for rustc 1.34?

> Do you know if there's any data on which rustc versions are still in use?

Nope.  It's probably possible to collect statistics from CI configurations of crates.io projects, but that would only give how many crates provides background compat to a given version, not how many end-users actually require it ...

ramosbugs:
> How about dropping support fore futures 0.1 and keeping synchronous-only support for rustc 1.34?

what problem does this solve? if the goal is to have the public interface use `http` 0.2 instead of 0.1, then removing `futures` 0.1 doesn't further that goal, but it does break compatibility with people whose codebases aren't on async/await yet...

kaj:
> what problem does this solve? if the goal is to have the public interface use `http` 0.2 instead of 0.1, then removing `futures` 0.1 doesn't further that goal, but it does break compatibility with people whose codebases aren't on async/await yet...

The goal here (as I see it) is not to remove any one version of any one library, but rather to have a set of dependencies as small and unconfusing as possible.

As futures 0.1 was always targeted at early adopters and explicitly an experiment before getting things standardised, I think very few of the codebases that aren't on asyc/await yet are using futures at all.

ramosbugs:
I put quite a bit of effort into [the docs](https://docs.rs/oauth2/3.0.0-alpha.9/oauth2/) that describe how to import this library, including the various combinations of feature flags. I don't think it significantly improves things to remove support for `futures` 0.1. Folks who don't need that dependency can disable it quite easily by reading the docs.

I've started working on an optional feature flag to add an `http` 0.2 compatibility layer for folks who want a custom HTTP client interface in which `HttpRequest` and `HttpResponse` use the `http` 0.2 types. I think that strikes the right balance of compatibility, complexity, and build time.

kaj:
I'm more and more conviced it would be best to drop all support for http 0.1, reqwest 0.9, and futures 0.1.  Keeping them was a good idea in june 2019, when the first alpha was released.  But since 3.0.0 is still in alpha, I think it is a good idea to drop the legacy support before cutting the actual 3.0.0 release rather than after.

PR #90, opened 2020-05-06
Replace deprecated "failure" with "thiserror" and "anyhow"

s-panferov:
The `failure` crate was deprecated recently, so this PR removes it
in favor of `thiserror`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -83,7 +83,7 @@
 //!    Synchronous HTTP clients should implement the following trait:
 //!    ```ignore
 //!    FnOnce(HttpRequest) -> Result<HttpResponse, RE>
-//!    where RE: failure::Fail
+//!    where RE: std::error::Error + Send + Sync + 'static 
 //!    ```
 //!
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -994,7 +994,7 @@ where
 
     fn prepare_request<RE>(self) -> Result<HttpRequest, RequestTokenError<RE, TE>>
     where
-        RE: Fail,
+        RE: Error,
```
s-panferov, line 997: Not sure if we have to add `Send + Sync + 'static` bounds here. `Fail` trait had them, but tests seem to work fine without them.

ramosbugs:
thanks for the PR, and apologies for the slow response. I do plan to merge this in, but I need to coordinate it with migrating the error handling in my employer's codebase, since this affects the `cause` chaining that differs between `failure` and the other error handling crates

ramosbugs:
Merged this in 09529db9bdc5059b7d0960b982486d42ea85f326

PR #92, opened 2020-06-01
Allows switching between native-tls and rustls-tls

mettke:
It would be awesome if one is able to switch between native and rustls. This is a first idea of how to archive that. 

Another idea would be to introduce the flags `native-tls` and `rustls-tls` similar to reqwest but those would always add both dependencies. 

Any thoughts on this? I would like to change this in [openidconnect-rs](https://github.com/ramosbugs/openidconnect-rs) too once we find a good solution

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,11 +15,13 @@ all-features = true
 maintenance = { status = "actively-developed" }
 
 [features]
-default = ["reqwest-09"]
+default = ["reqwest-09-native"]
 futures-01 = ["futures-0-1"]
 futures-03 = ["futures-0-3", "async-trait"]
-reqwest-09 = ["reqwest-0-9"]
-reqwest-010 = ["reqwest-0-10", "http-0-2"]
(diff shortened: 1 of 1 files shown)


mettke:
Ok I just figured out that it is not possible to propagate this change to openidconnect-rs as ring once again is a pain in the ass. Guess this cannot be merged before https://github.com/ramosbugs/openidconnect-rs/pull/4 or before ring is able to have multiple versions at once

ramosbugs:
hey, thanks for the PR! I think it's already possible for clients to use any desired TLS implementation (including for `reqwest`) by defining their own HTTP client (`HttpRequest -> Result<HttpResponse, RE>`) implementation. the bundled `reqwest` client is provided as a convenience.

Given that it's already possible to do this outside of the crate, I think it's probably not worth the additional complexity of 2 new feature flags plus a breaking change to 2 existing feature flags in order to add this functionality to the crate.

mettke:
That's a very good point. Thanks I did not think about that

PR #93, opened 2020-06-03
Updating to sha2 0.8

mettke:
sha2 got updated to v0.8. Updating works without any compile issues so there are no api changes

Update:
Sha2 is used only in letterboxd.rs and types.rs. Both do not use it in a user facing api.

The most notable change that happend in the sha2 crate is the update of digest to 0.8 and block-buffer to 0.7. They had to change their user facing api but not the parts this crate is using.


ramosbugs:
Thanks!

PR #95, opened 2020-06-06
Add ability to override the authorization url

mettke:
Adds method and test to override `redirect_url` on authorization.

Ref: ramosbugs/openidconnect-rs/issues/18

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -679,7 +679,7 @@ where
             client_id: &self.client_id,
             extra_params: Vec::new(),
             pkce_challenge: None,
-            redirect_url: self.redirect_url.as_ref(),
+            redirect_url: self.redirect_url.as_ref().map(Cow::Borrowed),
             response_type: "code".into(),
             scopes: Vec::new(),
             state: state_fn(),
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -852,6 +852,14 @@ impl<'a> AuthorizationRequest<'a> {
         self
     }
 
+    ///
+    /// Overrides the `redirect_url` to the one specified.
+    /// 
+    pub fn set_redirect_url(mut self, redirect_url: &'a RedirectUrl) -> Self {
```
ramosbugs, line 858: I think this lifetime may make the API tricky to use since clients will have to ensure that the `RedirectUrl` outlives the `AuthorizationRequest`. instead, let's change the type of `self.redirect_url` to be a `Cow<a', RedirectUrl>` and then we can pass an owned `RedirectUrl` here.

in the common case where users don't call `set_redirect_url`, we'll still avoid any extra clones, while supporting an easier API for this new use case.

mettke: I thought that you might say that. Will do

mettke:
I made the changes and they are ready for review again. I probably should have said something earlier

ramosbugs:
thanks! for some reason GitHub didn't send me a notification 🤔

PR #96, opened 2020-06-06
Updating to base64 0.12

mettke:
Hopefully the last update in this crate for a bit.

Changes:
https://github.com/marshallpierce/rust-base64/compare/v0.10.1...v0.12.1
The commits show that its mostly new stuff, modularization and the switch to no_std

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ reqwest-010 = ["reqwest-0-10", "http-0-2"]
 
 [dependencies]
 async-trait = { version = "0.1", optional = true }
-base64 = "0.10"
+base64 = "0.12"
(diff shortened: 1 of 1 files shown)


ramosbugs:
thanks! hopefully this won't run into the trailing bit issues you mentioned in ramosbugs/openidconnect-rs#21.

mettke:
Sorry... This is affected too as we are using URL Decoding too. Sorry I really should have left a comment. Will create a PR later today for oauth2-rs and openidconnect-rs to correct the config and use lax parsing too

mettke:
Nevermind I had a look and we are using base64 only to encode. Thus here is are no problems to be expected

PR #98, opened 2020-06-13
Updating to sha2 0.9

mettke:
Another easy update. I have, however, a few more things I would like to do:

* Update to 1.36 as min (its now a year old and would allow us to update unicode-normalization)
* Switch http1 and http2. Currently we have to use http1 to implement our own http clients. Switching would allow us to use http2. It means making http optional and http-0-2 the default. Of course it does introduce quite some changes to the user api.

Any chance those would get merged if I create PRs?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ reqwest-0-9 = { version = "0.9", optional = true, package = "reqwest" }
 reqwest-0-10 = { version = "0.10", optional = true, features = ["blocking"], package = "reqwest" }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
-sha2 = "0.8"
+sha2 = "0.9"
 # To support rustc < 1.36
 unicode-normalization = "=0.1.9"
 url = "2.1"


mettke:
Guess this one is part of having to update to 1.36

ramosbugs:
from https://github.com/ramosbugs/oauth2-rs/issues/79#issuecomment-577029003, my understanding was that `http` 0.2 requires 1.39 (~7 months old now). does it work with 1.36? of course, that comment was 6 months ago, so a version bump seems less problematic than it did a month after 1.39 was released.

I think at this point, it's probably ok to increase the minimum supported Rust version to 1.39 since this crate's 3.0 release is still in alpha and this shouldn't cause any significant semver issues (hopefully something like https://github.com/rust-lang/rust/issues/65262 will get merged to formalize tracking this).

aside from the Rust version issue, changing the external interface to `http` 2.0 would be fairly disruptive to current users. I think at this point it would make sense to cut a stable 3.0 beta branch and make the main development branch `4.0.0-alpha.1` so that users expect that level of disruption before trying to upgrade.

thoughts?

mettke:
I had a look at it and you're right. I does only compile with 1.39. 
Wow its really difficult to keep track of that. I really hope they find a way to add and somewhat enforce a Manifest key like that, it would make stuff quite a lot easier.

With that said, I think you're right. Making this version stable should be the way to go before we think of things that break the API. I used oauth-rs and openidconnect-rs for quite a while now and I think both are ready for a release. Their API is solid and the ability to use custom http clients is pretty neat. 

So in my opinion we can go ahead. If you, however, have a few more things you would like to do feel free and create a Tracking issue for v3 and I'm glad to help. Once we have a v4 version in master I would like to create a tracking issue for v4 so we can discussion what we want to do in the new version.

I will close this PR for know.

ramosbugs:
I've gone ahead and released 3.0.0 (which now lives in the `support/3.x` branch) and removed the CI tests for 1.34 from the `main` branch (which I also took this opportunity to rename from `master`). I'll make similar changes to `openidconnect` soon.

Issue #99, opened 2020-06-15
Release 4


mettke:
This is the tracking Issue for Release 4

### ToDo

- [x] Remove future-1 support
- [x] Remove reqwest 0.9 support
- [x] Switching from http1 to http2
- [x] Replace `rust-crypto`
- [x] Switch to rusttls by default #100 
- [x] Replace failure #90 
- [x] Switch to `default-features = false` in openidconnet-rs
- [x] Token Introspection #97

mettke:
#### Removing default features

Currently it is very difficult to remove the reqwest and thus a dependency on openssl when using openidconnect. The reason for this is, that oauth-rs has reqwest as default feature and openidconnect carries that default feature with it without the possibility of disabling it. 

Personally I would like to see a concept which is more like tokio which does not add anything by default but allows the user to specify what he/she wants.

#### Remove future-1 support

future-1 did not receive any update for 9 month now. Question is whether we want to continue supporting users who use it. 

#### Replace failure with concrete error type

As described in #90 failure got deprecated. I'm however not comfortable with having a crate like failure, thiserror or anyhow in a library crate. Those crates are, in my opinion, made for easy error handling in end user crates/binaries. We, however, should have a look whether it is possible to create a concrete error type which is flexible enough to handling unknown error.

mettke:
@ramosbugs I would like to hear your feedback before starting to work on this (sorry for the noisy if you just did not have time to get it to it - just making sure that you know)

ramosbugs:
> #### Removing default features
> Currently it is very difficult to remove the reqwest and thus a dependency on openssl when using openidconnect. The reason for this is, that oauth-rs has reqwest as default feature and openidconnect carries that default feature with it without the possibility of disabling it.

I'm confused about this. Why doesn't disabling the default features on `openidconnect` remove the dependency on `reqwest-09` and keep it from being built? https://github.com/ramosbugs/openidconnect-rs/blob/82aa3c6cf6ab39934e2cc64d5648282176444c26/Cargo.toml#L18

> Personally I would like to see a concept which is more like tokio which does not add anything by default but allows the user to specify what he/she wants.

I generally agree, but this crate isn't very useful out of the box without an HTTP client implementation, and the vast majority of users will probably want to use a simple, built-in HTTP client like the reqwest-based one this crate provides. As long as there's a straightforward way to disable the default, I think it's fine to continue building a client by default.

---

> #### Remove future-1 support
[...]

PR #100, opened 2020-06-15
Add rustls features

enfipy:
Better to add features to turn on rustls in `reqwest`.

Library `native-tls` is [dynamically linked](https://users.rust-lang.org/t/any-reasons-to-prefer-native-tls-over-rustls/37626) and it's hard sometimes to set up the environment for it. I added features that enable rustls in reqwest but by default `reqwest` uses `native-tls` in feature `default-tls` and needs to be disabled to make work rustls.

So to use rustls with oauth2 all you need to specify:
```rs
oauth2 = { version = "3.0", features = ["reqwest-010", "reqwest-010-rustls"], default-features = false }
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,9 @@ default = ["reqwest-09"]
 futures-01 = ["futures-0-1"]
 futures-03 = ["futures-0-3", "async-trait"]
 reqwest-09 = ["reqwest-0-9"]
+reqwest-09-rustls = ["reqwest-0-9/rustls-tls"]
 reqwest-010 = ["reqwest-0-10", "http-0-2"]
+reqwest-010-rustls = ["reqwest-0-10/rustls-tls"]
 
 [dependencies]
 async-trait = { version = "0.1", optional = true }
@@ -32,8 +34,8 @@ futures-0-3 = { version = "0.3", optional = true, package = "futures" }
 http = "0.1"
 http-0-2 = { version = "0.2", optional = true, package = "http" }
 rand = "0.7"
-reqwest-0-9 = { version = "0.9", optional = true, package = "reqwest" }
-reqwest-0-10 = { version = "0.10", optional = true, features = ["blocking"], package = "reqwest" }
+reqwest-0-9 = { version = "0.9", optional = true, default-features = false, package = "reqwest" }
+reqwest-0-10 = { version = "0.10", optional = true, default-features = false, features = ["blocking"], package = "reqwest" }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 sha2 = "0.8"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,13 +15,15 @@ all-features = true
 maintenance = { status = "actively-developed" }
 
 [features]
-default = ["reqwest-09"]
+default = ["reqwest-09", "reqwest-09-native-tls"]
 futures-01 = ["futures-0-1"]
 futures-03 = ["futures-0-3", "async-trait"]
 reqwest-09 = ["reqwest-0-9"]
 reqwest-09-rustls = ["reqwest-0-9/rustls-tls"]
+reqwest-09-native-tls = ["reqwest-0-9/default-tls"]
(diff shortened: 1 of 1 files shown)


mettke:
Thanks for the PR. However a similar one #92 was rejected recently to not further complicate the feature flags in this crate.

To get rid of the openssl dependency it is possible to specify your own HttpClient using `HttpRequest -> Result<HttpResponse, RE>` and disabling `default-features = false`.

enfipy:
> Thanks for the PR. However a similar one #92 was rejected recently to not further complicate the feature flags in this crate.
> 
> To get rid of the openssl dependency it is possible to specify your own HttpClient using `HttpRequest -> Result<HttpResponse, RE>` and disabling `default-features = false`.

Well, it's very unclear and more code to write. Because you need to specify another 2 libraries like `http` and `reqwest`. I think better to provide features like it done in your PR.

ramosbugs:
my main objection to #92 was the feature flag complexity that it introduced (since it had to work with both reqwest 0.9 and 0.10). however, I think we have an opportunity to revisit this now that we've moved on to a new major release (4.x), which is a good time to simplify things by removing legacy support (as @mettke alluded to in #99).

for 4.x, it seems reasonable to remove support for both futures 0.1 and reqwest 0.9 and leave only (optional) futures 0.3 and reqwest 0.10 support. as part of this change, how about we change the reqwest 0.10 import to unconditionally use `rustls-tls` instead of `default-tls`? I think it makes sense to drop support for `default-tls` given that:
1. `native_tls` causes frequent build issues (I've seen many coworkers waste time dealing with OpenSSL linking/versioning issues)
2. many Linux LTS releases (e.g., Ubuntu 14.04 and 16.04) don't have OpenSSL package versions newer than 1.0.2 available, which runs into both security issues and known certificate verification issues [[1](https://twitter.com/sleevi_/status/1266647545675210753?s=20)] [[2](https://twitter.com/sleevi_/status/1272763292168196096)]

so, at a minimum, we probably want to change the default to `rustls-tls`, but using `default-tls` on purpose at this point seems so ill-advised that we can just ask people to supply their own `HttpRequest -> HttpResponse` implementation if they truly want it (in effect, inverting the current burden this crate imposes on `rusttls-tls` users). most users probably won't care and won't even notice the change in the underlying TLS implementation.

enfipy:
> my main objection to #92 was the feature flag complexity that it introduced (since it had to work with both reqwest 0.9 and 0.10). however, I think we have an opportunity to revisit this now that we've moved on to a new major release (4.x), which is a good time to simplify things by removing legacy support (as @mettke alluded to in #99).
> 
> for 4.x, it seems reasonable to remove support for both futures 0.1 and reqwest 0.9 and leave only (optional) futures 0.3 and reqwest 0.10 support. as part of this change, how about we change the reqwest 0.10 import to unconditionally use `rustls-tls` instead of `default-tls`? I think it makes sense to drop support for `default-tls` given that:
> 1. `native_tls` causes frequent build issues (I've seen many coworkers waste time dealing with OpenSSL linking/versioning issues)
> 2. many Linux LTS releases (e.g., Ubuntu 14.04 and 16.04) don't have OpenSSL package versions newer than 1.0.2 available, which runs into both security issues and known certificate verification issues [[1](https://twitter.com/sleevi_/status/1266647545675210753?s=20)] [[2](https://twitter.com/sleevi_/status/1272763292168196096)]
> 
> so, at a minimum, we probably want to change the default to `rustls-tls`, but using `default-tls` on purpose at this point seems so ill-advised that we can just ask people to supply their own `HttpRequest -> HttpResponse` implementation if they truly want it (in effect, inverting the current burden this crate imposes on `rusttls-tls` users). most users probably won't care and won't even notice the change in the underlying TLS implementation.

So I can close this PR and you will add it in new v4, right?

ramosbugs:
> So I can close this PR and you will add it in new v4, right?

I'm open to PRs that implement the changes I suggested, but I can't guarantee when I'll have the time to make these changes myself.

mettke:
@enfipy Yes feel free to close it. I will add this to our roadmap in #99

PR #102, opened 2020-06-19
Removing future 0.1 and reqwest 0.9

mettke:
So here it is, removing future 0.1 and reqwest 0.9.
I know its a bit much to review (332 changes) but most is only removing stuff.

#99. 

## Changes 

* future 0.1 and reqwest 0.9 are both pretty old
  and should not be used anymore. Thus we switch to
  only supporting future 0.3 and reqwest 0.10
* unicode-normalization is unused and thus removed
* tokio was used inside future 0.1 and now got removed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,25 +6,16 @@ rust:
 sudo: false
 #cache: cargo
 before_script:
-  - cargo install --force cargo-when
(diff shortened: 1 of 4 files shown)


ramosbugs:
awesome, thanks! deleting stuff is fun :-)

PR #104, opened 2020-06-19
RUSTSEC-2016-0005: rust-crypto unmaintained

mettke:
rust-crypto is unmaintained and should not be used in example code anymore.
Also updating sha2 which should be its own PR but I do not want to create to many at once


ramosbugs:
thanks!

Issue #105, opened 2020-06-19
Device Flow/Device Authorization Grant Support


smolck:
Is it possible to do this with this library? If not, could support for this flow be added?

https://oauth.net/2/device-flow/

ramosbugs:
The device flow isn't currently supported, but it would be a nice enhancement to this crate.

Adding support would require adding methods for both the Device Authorization request and the Device Access Token request. The former uses a new endpoint and can be supported by following similar steps to the ones I outlined in #97 for the Token Introspection endpoint. The Device Access Token request reuses the existing token endpoint, so we just need to add a `Client::exchange_device_code` method.

I'd happily accept a PR that does this (with corresponding test coverage).

maxdymond:
FWIW I'm working on this at the moment. So far the interface is like this (running against a Google OAuth2 application)

```rust
    // Set up the config for the Google OAuth2 process.
    let device_client = BasicClient::new(
        google_client_id.clone(),
        Some(google_client_secret.clone()),
        auth_url.clone(),
        Some(token_url.clone()),
    )
    .set_device_authorization_url(device_auth_url);

    // Request the set of codes from the Device Authorization endpoint.
    let details = device_client
        .exchange_device_code()
        .add_scope(Scope::new("profile".to_string()))
        .request_codes(http_client)
        .expect("Failed to request codes from device auth endpoint");

    // Display the URL and user-code.
    println!(
        "Open this URL in your browser:\n{}\nand enter the code: {}",
        details.verification_uri().to_string(),
        details.user_code().to_string()
    );

    // Now poll for the token
    let token = BasicClient::new(
        google_client_id,
        Some(google_client_secret),
        auth_url,
        Some(token_url),
    )
    .set_device_authorization_details(details)
    .exchange_device_access_token()
    .request(http_client)
    .expect("Failed to request access token from token auth endpoint");

    println!("Google returned the following token:\n{:?}\n", token);
```

I've written the first half (up to and including the displaying of codes to the user), and now I'm working on how to repeatedly poll the endpoint.

Issue #97, opened 2020-06-09
OAuth 2.0 Token Introspection (RFC 7662)


jeroenvervaeke:
I'm in the need for OAuth 2.0 token introspection ([RFC 7662](https://tools.ietf.org/html/rfc7662)).

> This specification defines a method for a protected resource to query
   an OAuth 2.0 authorization server to determine the active state of an
   OAuth 2.0 token and to determine meta-information about this token.
   OAuth 2.0 deployments can use this method to convey information about
   the authorization context of the token from the authorization server
   to the protected resource.

_I want to implement this feature_ but I'm not sure if you guys feel like it belongs here. Since the readme states that this library is an `strongly-typed implementation of OAuth2 (RFC 6749)`.
I also realize there is a library available already ([tokkit](https://github.com/chridou/tokkit)), but personally I feel like it would be more convenient to have all these features in one library.

What do you guys think. Should this RFC be added to this library? Should it be added by default? Should this feature be added behind a feature flag etc?

Thanks in advance!

ramosbugs:
Hey, this seems like a reasonable feature to incorporate into the library. There's some precedent for including closely-related RFCs like PKCE directly into this crate.

I think the inclusion criteria should be that:
1. the functionality is of interest to many users of this crate, and
2. it doesn't significantly increase the complexity of the crate.

In this case, I think both criteria apply, so I'd be open to incorporating it. By contrast, something like OpenID Connect is far more complex, so I put that into a separate crate (https://github.com/ramosbugs/openidconnect-rs) built on top of this one.

For clarity, I think it makes sense to incorporate the token introspection client request/response from that RFC, but not any functionality that directly introspects into JWTs or anything like that.

From glancing at the RFC, I think we can incorporate this into the crate most cleanly by:
1) adding a `Client::set_introspection_url` method for setting the URL (which can be leveraged by multiple subsequent introspection calls, similar to the pattern for setting the `auth_url` and `token_url` in `Client::new`, but without introducing a breaking change to that method)
2) adding a `Client::introspect_token` method that follows the same pattern as `exchange_code`, etc. (i.e., it should return an `IntrospectionRequest` struct).
3) defining a strongly-typed `IntrospectionResponse` struct leveraging the NewType pattern where appropriate
4) adding appropriate unit tests

jeroenvervaeke:
I finally find some time to implement this and create a PR (See #117)
[...]

Issue #80, opened 2019-12-29
async actix-web clients fail to build


ensc:
When trying to build an actix-web based client for the http requiest, I get errors about missing `Send` markers. E.g. a stripped down example might be

```rust
async fn request()
{
    let mut response = awc::ClientBuilder::new()
	.finish()
	.get("")
	.send_body("")
	.await
	.unwrap();

    response.body().await.unwrap();
}

async fn request_async<FN, F>(f: FN) -> ()
where
    FN: FnOnce() -> F + Send,
    F:  std::future::Future<Output = ()> + Send,  /// <<<< removing this Send makes it work
{
    f().await
}

async fn auth()
{
    request_async(request).await;
}

fn main() {
}
```

which results into

```
error[E0277]: `std::rc::Rc<awc::ClientConfig>` cannot be sent between threads safely
   = help: within `impl std::future::Future`, the trait `std::marker::Send` is not implemented for `std::rc::Rc<awc::ClientConfig>`
```

(and much more).

The `request_auth()` function corresponds to oauth2:'s methods which have a similar signature.

When removing the `Send` requirement from `F`, things seem to be fine.  In oauth2 this seems to be required due to `#[async_trait]` but accordingly its [documentation](https://docs.rs/async-trait/0.1.22/async_trait/#non-threadsafe-futures) this can be explicitly disabled by writing

```rust
#[async_trait(?Send)]
```

Is there a reason why `Send` is required for the `http_client` or would it be possible to avoid it (which makes writing awc clients much easier)?

rust-1.40.0
awc-1.0.1
actix-web-2.0 (futures-0.3)
oauth2-3.0.0-alpha.7

ramosbugs:
Hmm where does the `Rc` come from? wouldn't making it an `Arc` solve the problem?

I don't fully understand what a "non-threadsafe future" would be... that seems somewhat contradictory, no?

ensc:
The `Rc` is coming from actix and there are yet more more objects which are not thread safe.  There is probably more work required than replacing `Rc` by `Arc` to change this.

I have no idea whether a non-threadsafe future would be contradictory. The `async` functions (the `rust` language feature) do not have the `Send` requirement; it was explicitly added by the `async_trait` crate (with a way to opt-out).  

Perhaps the executor sticks such futures to a thread while `Send` futures can be assigned to other ones.

asonix:
[...]

PR #106, opened 2020-06-23
Switching reqwest to rustls

mettke:
I think with this one in, we may release the first alpha version so that we are able to continue on openidconnect-rs. 

Alternatively we may thing about releasing as there are no more api updates expected and #97 and #105 would only add new stuff (minor version bump)

Edit: before releasing we may want to take a closer look at #80. Because if we do decide to remove async_trait it would introduce a breaking change

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ thiserror="1.0"
 futures-0-3 = { version = "0.3", optional = true, package = "futures" }
 http = "0.2"
 rand = "0.7"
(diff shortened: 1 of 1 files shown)


ramosbugs:
thanks!

as far as releases, I view alpha releases as non-stable APIs, so it's fine to continue making breaking changes until the first beta release (though it's of course best to minimize breakages whenever possible, so it would be nice to get #80 fixed). I agree that #97 and #105 shouldn't require breaking changes in any case, so there's no need to hold up a release for those features.

PR #108, opened 2020-06-27
Removing async_trait and refactoring

mettke:
It took me a bit but now its done. One of the bigger commits which removes the future crate (replaced wit std::future::Future; both are interchangeable and thus should not result in changes when calling the api) and the async_trait. 

Also combined a few modules to make reqwest look more like curl and moved the stuff from async_internals to the other functions in lib.rs. 

(Btw I'm now the the second highest contributor when it comes to deletions. I'm not sure whether that is a good thing :laughing:)
Edit: Forget to mention #80


ramosbugs:
thanks! added some additional cleanup in 3cfde6409ad9d1ba71cc3fa5c4a426ee17e0744e

PR #109, opened 2020-06-27
Remove send and sync

mettke:
Followup of #108 

We may want to consider removing Send and Sync requirements completely. They are not requires as they are added when the generic type resolves to a send and sync type. 

The only thing we have to make sure is that all our types are Send and Sync. I did that by adding a Test which checks exactly that. I also added another example which shows the github example using async (Prove that it still works using multithreaded executor)


src/tests.rs
```diff
@@ -1455,3 +1455,84 @@ fn test_secret_redaction() {
     let secret = ClientSecret::new("top_secret".to_string());
     assert_eq!("ClientSecret([redacted])", format!("{:?}", secret));
 }
+
+#[test]
```
ramosbugs, line 1459: nice :-)

PR #110, opened 2020-07-07
Enable Serde for re-exported Url

phated:
I'm not sure if this is needed, but I was using the re-exported `url::Url` and expected it to have serde support. I figured I would open this in case it made sense to add. Cheers!


ramosbugs:
Thanks!

Issue #107, opened 2020-06-25
Derive Eq, Hash for `CsrfToken`


jhpratt:
Can `#[derive(PartialEq, Eq, Hash)]` be added to `CsrfToken`? This would allow usage within `HashSet` and `BTreeSet`, which would let me keep track of outstanding tokens. The only alternative I can think of is to use a `Vec`, but that's O(n), which obviously isn't ideal.

ramosbugs:
These traits are intentionally omitted from the types that wrap secrets in order to prevent [timing side channels](https://en.wikipedia.org/wiki/Timing_attack). To safely use CSRF tokens and other secrets as lookups in data structures, I recommend creating another wrapper type that _only_ performs comparisons on a collision- and preimage-resistant hash (e.g., SHA-2) of the underlying secret and never compares the raw secrets themselves.

In theory we could implement the traits above directly on `CsrfToken` and the other secret types using a SHA-2 variant, but I think it would surprise users (i.e., violate the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)) to implement those traits in a computationally-expensive manner. I would be open to a PR that adds those traits behind a feature flag (e.g., `timing-resistant-secret-traits`), though.

jhpratt:
Using a dedicated crate (like sha2, as mentioned) to hash and compare seems reasonable, and wouldn't require much effort on my end.

Timing attacks are always one thing I never think about — which is why it's best to leave it to others!

Thank you for the quick response. I'll close this, as I won't have a ton of time to put together a proper pull request.

PR #111, opened 2020-07-07
Add derive(Eq, PartialEq) to CsrfToken

phated:
I find myself wanting to compare `state == expected_state` with them both being a `CsrfToken`. Do you think it makes sense to derive Eq and PartialEq for this type?

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -496,7 +496,7 @@ new_secret_type![
     /// via the `state` parameter.
     ///
     #[must_use]
-    #[derive(Clone, Deserialize, Serialize)]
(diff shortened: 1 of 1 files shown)


ramosbugs:
I felt it made sense to omit these to avoid timing side channels; see #107 for a more complete rationale

phated:
Ha, I actually had that as a TODO in my application code. Makes sense.

phated:
@ramosbugs it would be cool if things that made it easier but unsafe could be enabled with a "playground" or "unsafe" feature flag. I know a few people that have run into this same issue as me.

PR #115, opened 2020-09-29
Add "plain" pkce mode, refactor sha256 to share code

ccope:
I implemented this to talk to a third-party API that doesn't support any PKCE challenge types besides "plain". I split out PkceCodeVerifier creation from the sha256 functions so it can be shared with other challenge type implementations. I'm pretty new to rust, so please let me know if there was a better way to achieve this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ maintenance = { status = "actively-developed" }
 
 [features]
 default = ["reqwest-010"]
+pkce-plain = []
 reqwest-010 = ["reqwest-0-10"]
 
 [dependencies]
--- a/src/types.rs
+++ b/src/types.rs
@@ -428,7 +428,7 @@ impl PkceCodeChallenge {
     }
 
     ///
-    /// Generate a new random, base64-encoded SHA-256 PKCE code.
+    /// Generate a new random, base64-encoded SHA-256 PKCE challenge code and verifier.
     ///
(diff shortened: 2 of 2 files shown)


ramosbugs (commented):
thanks for the PR!

I originally left out support for `plain` since [RFC 7636](https://tools.ietf.org/html/rfc7636#section-7.2) says (emphasis verbatim):
> "plain" SHOULD NOT be used and exists only for
   compatibility with deployed implementations where the request path is
   already protected.  The "plain" method SHOULD NOT be used in new
   implementations, unless they cannot support "S256" for some technical
   reason.

It's hard for me to envision a compelling technical reason for a server not to support S256 when any server that can handle TLS and an OAuth flow clearly has enough CPU to compute a SHA-256 hash. Consequently, I'm willing to merge in support for `plain`, but I'd prefer to do it behind a feature flag since it really shouldn't be used in nearly all cases, and this library aims to provide APIs that help users follow best practices without needing to read the RFCs.

Mind adding a `pkce-plain` (non-default) Cargo feature flag? Then we can guard this code with `#[cfg(feature = "pkce-plain")]`.

Also, looks like [CI failed](https://travis-ci.org/github/ramosbugs/oauth2-rs/jobs/731421988) due to `cargo fmt`.

ramosbugs:
thanks!

PR #117, opened 2020-11-01
Implemented OAuth 2.0 Token Introspection (RFC 7662)

jeroenvervaeke:
Implemented #97 

> 1. adding a Client::set_introspection_url method for setting the URL (which can be leveraged by multiple subsequent introspection calls, similar to the pattern for setting the auth_url  xand token_url in Client::new, but without introducing a breaking change to that method) :heavy_check_mark: 
> 2. adding a Client::introspect_token method that follows the same pattern as exchange_code, etc. (i.e., it should return an IntrospectionRequest struct). :heavy_check_mark: 
> 3.  defining a strongly-typed IntrospectionResponse struct leveraging the NewType pattern where appropriate :heavy_check_mark: 
> 4.  adding appropriate unit tests :heavy_check_mark:


Emilgardis:
What's the status on this? CI errors seem spurious

ramosbugs:
hey, apologies for the delay! I've been wrapped up in some other projects.

I'll take a look at this PR on Wed or Thurs of this week. in the meantime, you should be able to use a Cargo git dependency on the author's branch if you need the functionality sooner than I'm able to merge it.

Emilgardis:
Yes of course, I'm just more excited for a new release so I can pin on a stable release instead ;)

src/lib.rs
```diff
@@ -1388,6 +1424,140 @@ where
...
+    where
+        RE: Error + 'static,
+    {
+        let mut params: Vec<(&str, &str)> = vec![("token", self.token.secret())];
+        if let Some(ref token_type_hint) = self.token_type_hint {
+            params.push(("token_type_hint", token_type_hint));
+        }
+
+        Ok(endpoint_request(
+            self.auth_type,
```
ramosbugs, line 1515: the spec leaves the details around client auth somewhat open-ended:
>  To prevent token scanning attacks, the endpoint MUST also require
   some form of authorization to access this endpoint, such as client
   authentication as described in OAuth 2.0 [RFC6749] or a separate
   OAuth 2.0 access token such as the bearer token described in OAuth
   2.0 Bearer Token Usage [RFC6750].  The methods of managing and
   validating these authentication credentials are out of scope of this
   specification.

it would probably be worth adding a `set_auth_method` (or similar) function to choose between:
1. client id+secret auth (either in the `Authorization` header or request body, as determined by `self.auth-type`)
2. specifying a bearer token (see the first example in [Sec. 2.1](https://tools.ietf.org/html/rfc7662#section-2.1))
3. no automatic auth added by this crate (clients can either add their own custom headers by using a custom `http_client`, or we could add a corresponding `add_request_header` function similar to `add_extra_param`)

however, using client id+secret auth seems like a sensible default, and we should be able to add `set_auth_method` in the future without introducing a breaking change, so it's not a requirement for this PR.

src/lib.rs
```diff
@@ -2096,6 +2266,324 @@ where
     }
 }
 
+///
```
ramosbugs, line 2269: thanks for the thorough docs!

src/lib.rs
```diff
@@ -2096,6 +2266,324 @@ where
...
+    ///
+    /// OPTIONAL.  Client identifier for the OAuth 2.0 client that
+    /// requested this token.
+    ///
+    fn client_id(&self) -> Option<&ClientId>;
+    ///
+    /// OPTIONAL.  Human-readable identifier for the resource owner who
+    /// authorized this token.
+    ///
+    fn username(&self) -> Option<&String>;
```
ramosbugs, line 2312: let's return `&str` instead of `&String` for these, which is a bit more idiomatic in Rust I think. I used `&String` in a few places from when I was less familiar with the Rust conventions

src/lib.rs
```diff
@@ -2096,6 +2266,324 @@ where
...
+    /// 2.0 [RFC6749].
+    /// Value is case insensitive and deserialized to the generic `TokenType` parameter.
+    ///
+    fn token_type(&self) -> Option<&TT>;
+    ///
+    /// OPTIONAL.  Integer timestamp, measured in the number of seconds
+    /// since January 1 1970 UTC, indicating when this token will expire,
+    /// as defined in JWT [RFC7519](https://tools.ietf.org/html/rfc7519).
+    ///
+    fn exp(&self) -> Option<&DateTime<Utc>>;
```
ramosbugs, line 2324: it looks like [`DateTime` implements `Copy`](https://docs.rs/chrono/0.4.19/chrono/struct.DateTime.html#impl-Copy) so I think we can return an owned copy instead of a reference for these

src/lib.rs
```diff
@@ -2096,6 +2266,324 @@ where
...
+where
+    EF: ExtraTokenFields,
+    TT: TokenType + 'static,
+{
+    active: bool,
+    #[serde(rename = "scope")]
+    #[serde(deserialize_with = "helpers::deserialize_space_delimited_vec")]
+    #[serde(serialize_with = "helpers::serialize_space_delimited_vec")]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(default = "none_field")]
```
ramosbugs, line 2379: I'm pretty sure serde does this by default for all `Option` fields, so I think we can remove all of these `default = "none_field"` attributes

src/tests.rs
```diff
@@ -1474,6 +1477,156 @@ fn test_error_response_serializer() {
     );
 }
 
+#[derive(Deserialize, Debug, Clone)]
+pub struct ObjectWithOpionalStringOrVecString {
```
ramosbugs, line 1481: nit: `Optional`

ramosbugs (commented):
thanks for the PR! just a few minor comments :-)

src/lib.rs
jeroenvervaeke: I started implementing but then I saw this library already used the `AuthType`, so I re-used that.
AuthType could indeed be extended with a bearer token option in a later phase

src/lib.rs
jeroenvervaeke: I thought so too, but as far as I remember it did not work without the attribute.
I'll verify it when I update this PR.

jeroenvervaeke:
Thanks for the response, I'll send an update in the next days

src/lib.rs
ramosbugs: I had been thinking that `AuthType` should only support the auth methods used by the token endpoint, but checking [RFC 6749](https://tools.ietf.org/html/rfc6749#section-2.3) again, I see:
> The authorization server MAY accept any form of client authentication meeting its security requirements.

given this, I'm fine with just using `AuthType` and extending it in the future. I pushed 3ccf4252da1ab81c80074296347a762456174132 to make it `#[non_exhaustive]`.

src/lib.rs
jeroenvervaeke: I verified, I was able to remove `#[serde(default = "none_field")]` from fields like `username: Option<String>` however, when the field has other attributes like `#[serde(deserialize_with` or `#[serde(with = ` the attribute seems to be required.
If you remove them the unit tests fail and complain about the field being missing.

src/lib.rs
```diff
@@ -506,7 +507,7 @@ pub enum AuthType {
 /// Stores the configuration for an OAuth2 client.
 ///
 #[derive(Clone, Debug)]
-pub struct Client<TE, TR, TT>
+pub struct Client<TE, TR, TT, TIR>
```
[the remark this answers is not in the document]
jeroenvervaeke: Updated

src/lib.rs
jeroenvervaeke: I've updated all but `fn aud(&self) -> Option<&Vec<String>>;` since it stores all audience strings as a string internally

src/lib.rs
jeroenvervaeke: Updated

src/tests.rs
jeroenvervaeke: Updated

jeroenvervaeke:
The plan to send an update in the following days got delayed quiet a bit.
I've updated my PR

ramosbugs:
thanks!

PR #118, opened 2021-01-05
Bump to `tokio 1.0` and `reqwest 0.11`

jtgeibel:
This drops the feature `reqwest-010` and replaces it with `reqwest-011` based on the new `reqwest 0.11` release. The version of tokio used in tests is also bumped to the stable 1.0 release.

cc #99

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,9 +15,9 @@ all-features = true
 maintenance = { status = "actively-developed" }
 
 [features]
-default = ["reqwest-010"]
+default = ["reqwest-011"]
 pkce-plain = []
-reqwest-010 = ["reqwest-0-10"]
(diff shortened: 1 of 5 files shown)


jtgeibel:
The test failure is because tokio has bumped it's minimum supported rustc to 1.45. Their [release announcement](https://tokio.rs/blog/2020-12-tokio-1-0) describes their MSRV as:

> Tokio will keep a rolling MSRV (minimum supported rust version) policy of 6 months. When increasing the MSRV, the new Rust version must have been released at least six months ago.

ramosbugs:
thanks! I'll merge this once I get CI to pass. I just pushed edad9ba6072e69a439952ce8ed46e27b61539299 to bump the MSRV and fix an unrelated CI issue

PR #120, opened 2021-02-07
wasm32 support

DusterTheFirst:
since reqwest already supports wasm32 targets, and all dependencies of this crate are already platform-agnostic, by disabling the blocking client implementation (which does not exist on wasm platforms), the crate can be used in wasm.
The curl backend, obviously, does not work in wasm targets, so a gate has been added to prevent its compilation with a simple error message for the user. The curl dependency has been made conditional so that it does not get compiled for wasm targets, preventing a wall of error messages hiding the error from this crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,6 @@ pkce-plain = []
 
 [dependencies]
 base64 = "0.12"
-curl = { version = "0.4.0", optional = true }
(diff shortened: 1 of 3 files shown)


ramosbugs:
thanks for the PR! I'll attempt to add a wasm32 target to CI so that this doesn't break down the road

PR #121, opened 2021-02-07
Add ureq http_client

### LINKED ISSUE #119: Add ureq http provider
Hello,

I've recently started using oauth2-rs for my application (and it works nicely), but I'm using ureq instead of curl to cut the dependency on that. Would you be interested in me upstreaming that or should users just implement the `http_client` method themselves for ureq (it's only like ~40 LOC)?

Cogitri:
fixes #119


Cogitri:
Tested authentication w/ Google with this

ramosbugs:
thanks!

PR #37, opened 2018-03-30
Deprecate set_state() and recommend an upgrade to 2.0

jtgeibel:
This PR deprecates the set_state() method in the 1.x series and recommends that the developer upgrade to 2.0.  I've bumped the version number to 1.3.1.

PR #122, opened 2021-02-09
RFC 7009 OAuth 2.0 Token Revocation

ximon18:
Initial DRAFT implementation intended for discussion and refinement.

Supports extensible token types for revocation needed to comply with [RFC 7009 Section 2.1](https://tools.ietf.org/html/rfc7009#section-2.1) support for custom token types.

Supports extensible response handling because RFC 7009 Section 2.1 says _"The content of the response body is ignored by the client as all necessary information is conveyed in the response code"_ but Serde requires some response definition. In testing with Google an empty JSON response was returned. If an alternate provider were to return something more in the body this would break the Serde parsing. However, this still assumes that the response is JSON which RFC 7009 does not appear to mandate.

Known issues:
- Breaks existing code that defines a custom client due to the need to extend the custom client type definition to include new types for revocable token and revocation error response. Note: the PR that added token introspection support had the same limitation.
- Only "tested" against Google. The Google API responds with success but this is not proof that the token was actually revoked. 
- Only "tested" with an access token as in initial tests Google did not issue a refresh token.
- Only the "happy" flow has been tested, i.e. correct handling of an HTTP 400 `unsupported_token_type` response has not been tested.
- Documentation comments are missing.
- I'm not sure that I've used the notion of oauth2 crate Basic and Standard types correctly.
- Perhaps a better approach would be to actually ignore the success resposne body as stated in RFC 7009 Section 2.1?
- The `EndpointError` idea mentioned in [openid-connect issue #37](https://github.com/ramosbugs/openidconnect-rs/issues/37) has not been implemented yet.

### DIFF
--- a/examples/google.rs
+++ b/examples/google.rs
@@ -13,11 +13,7 @@
 //! ...and follow the instructions.
 //!
 
-use oauth2::{
-    basic::BasicClient,
-    revocation::{StandardRevocableToken, StandardRevocationResponse},
-    TokenResponse,
-};
+use oauth2::{basic::BasicClient, revocation::StandardRevocableToken, TokenResponse};
 // Alternatively, this can be oauth2::curl::http_client or a custom.
 use oauth2::reqwest::http_client;
 use oauth2::{
@@ -148,23 +144,17 @@ fn main() {
 
             // Revoke the obtained token
             let token_response = token_response.unwrap();
-            let token_to_revoke = if let Some(token) = token_response.refresh_token() {
-                StandardRevocableToken::RefreshToken(token.clone())
-            } else {
-                StandardRevocableToken::AccessToken(token_response.access_token().clone())
+            let token_to_revoke: StandardRevocableToken = match token_response.refresh_token() {
+                Some(token) => token.into(),
+                None => token_response.access_token().into(),
             };
 
-            let revocation_response: StandardRevocationResponse = client
+            client
                 .revoke_token(token_to_revoke)
                 .request(http_client)
                 .expect("Failed to revoke token");
 
-            println!(
-                "Google returned the following revocation response:\n{:?}\n",
-                revocation_response
-            );
-
-            // The server will terminate itself after collecting the first code.
+            // The server will terminate itself after revoking the token.
             break;
         }
     }
--- a/examples/google.rs
+++ b/examples/google.rs
@@ -13,12 +13,12 @@
 //! ...and follow the instructions.
 //!
 
-use oauth2::basic::BasicClient;
+use oauth2::{TokenResponse, basic::BasicClient, revocation::{StandardRevocableToken, StandardRevocationResponse}};
 // Alternatively, this can be oauth2::curl::http_client or a custom.
 use oauth2::reqwest::http_client;
 use oauth2::{
     AuthUrl, AuthorizationCode, ClientId, ClientSecret, CsrfToken, PkceCodeChallenge, RedirectUrl,
-    Scope, TokenUrl,
+    RevocationUrl, Scope, TokenUrl,
 };
 use std::env;
 use std::io::{BufRead, BufReader, Write};
@@ -49,6 +49,10 @@ fn main() {
     // See below for the server implementation.
     .set_redirect_url(
         RedirectUrl::new("http://localhost:8080".to_string()).expect("Invalid redirect URL"),
+    )
+    // Google supports OAuth 2.0 Token Revocation (RFC-7009)
+    .set_revocation_url(
+        RevocationUrl::new("https://oauth2.googleapis.com/revoke".to_string()).expect("Invalid revocation endpoint URL"),
     );
 
     // Google supports Proof Key for Code Exchange (PKCE - https://oauth.net/2/pkce/).
@@ -127,12 +131,27 @@ fn main() {
             );
 
             // Exchange the code with a token.
-            let token = client
+            let token_response = client
                 .exchange_code(code)
                 .set_pkce_verifier(pkce_code_verifier)
                 .request(http_client);
 
-            println!("Google returned the following token:\n{:?}\n", token);
+            println!("Google returned the following token:\n{:?}\n", token_response);
+
+            // Revoke the obtained token
+            let token_response = token_response.unwrap();
+            let token_to_revoke = if let Some(token) = token_response.refresh_token() {
+                StandardRevocableToken::RefreshToken(token.clone())
+            } else {
+                StandardRevocableToken::AccessToken(token_response.access_token().clone())
+            };
+
+            let revocation_response: StandardRevocationResponse = client
+                .revoke_token(token_to_revoke)
+                .request(http_client)
+                .expect("Failed to revoke token");
+
+            println!("Google returned the following revocation response:\n{:?}\n", revocation_response);
 
             // The server will terminate itself after collecting the first code.
             break;
--- a/examples/google.rs
+++ b/examples/google.rs
@@ -13,7 +13,11 @@
 //! ...and follow the instructions.
 //!
 
-use oauth2::{TokenResponse, basic::BasicClient, revocation::{StandardRevocableToken, StandardRevocationResponse}};
+use oauth2::{
+    basic::BasicClient,
+    revocation::{StandardRevocableToken, StandardRevocationResponse},
+    TokenResponse,
+};
 // Alternatively, this can be oauth2::curl::http_client or a custom.
 use oauth2::reqwest::http_client;
 use oauth2::{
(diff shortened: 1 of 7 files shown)


ximon18:
@ramosbugs: the CI build only appears to be failing because of missing documentation comments. Before I invest more in this can you see if the basic approach seems okay? For example I'm not sure about the revocable token type matching and cloning or the way of handling the unknown ignorable response body.

src/revocation.rs
```diff
@@ -0,0 +1,163 @@
...
+use crate::{AccessToken, RefreshToken};
+
+pub trait RevocableToken {
+    fn secret(&self) -> &String;
+
+    fn token_type_hint(&self) -> &'static str;
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub enum StandardRevocableToken {
```
ramosbugs, line 15: let's make this `#[non_exhaustive]` so that the crate can add more token types in the future without breaking backward compatibility

src/lib.rs
```diff
@@ -523,12 +531,14 @@ pub enum AuthType {
...
 #[derive(Clone, Debug)]
-pub struct Client<TE, TR, TT, TIR>
+pub struct Client<RT, TE, TR, TT, TIR, RER>
 where
+    RT: RevocableToken,
     TE: ErrorResponse,
     TR: TokenResponse<TT>,
     TT: TokenType,
     TIR: TokenIntrospectionResponse<TT>,
+    RER: ErrorResponse,
```
ramosbugs, line 541: now that we have two `ErrorResponse` type parameters, it's probably time to document them in the doc comment. these were probably self-evident before, but less so now

src/revocation.rs
```diff
@@ -0,0 +1,163 @@
...
+    AccessToken(AccessToken),
+    RefreshToken(RefreshToken),
+}
+impl RevocableToken for StandardRevocableToken {
+    fn secret(&self) -> &String {
+        match self {
```
ramosbugs, line 21: > I'm not sure about the revocable token type matching and cloning

lgtm!

examples/google.rs
```diff
@@ -127,12 +136,33 @@ fn main() {
...
-            println!("Google returned the following token:\n{:?}\n", token);
+            println!(
+                "Google returned the following token:\n{:?}\n",
+                token_response
+            );
+
+            // Revoke the obtained token
+            let token_response = token_response.unwrap();
+            let token_to_revoke = if let Some(token) = token_response.refresh_token() {
+                StandardRevocableToken::RefreshToken(token.clone())
```
ramosbugs, line 152: is this the cloning you were asking about? I think it's fine since it's just one token per request, and the alternative would add some lifetime complexity

ramosbugs (commented):
thanks!

src/revocation.rs
```diff
@@ -0,0 +1,163 @@
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use std::fmt::Error as FormatterError;
+use std::fmt::{Debug, Display, Formatter};
+
+use crate::{basic::BasicErrorResponseType, ErrorResponseType, StandardErrorResponse};
+use crate::{AccessToken, RefreshToken};
+
+pub trait RevocableToken {
+    fn secret(&self) -> &String;
```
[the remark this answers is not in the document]
ximon18: Indeed, I just copied the interface of secret types.

src/revocation.rs
ximon18: Ah good idea.

src/lib.rs
ximon18: I wondered if it would be better to not even put the additional types at the struct level but only on the revoke fn, that way you don't break existing code that already has custom `Client`s... ?

examples/google.rs
ximon18: Yes

src/revocation.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    fn token_type_hint(&self) -> &'static str;
```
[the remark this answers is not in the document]
ximon18: I was thinking that as the token types have to be codified into an enum the actual type names are constant and known in advance and are unchanging. It is indeed more restrictive, and perhaps unnecessary, but it seemed ... right?

src/lib.rs
```diff
@@ -1575,6 +1621,116 @@ where
...
+        params.push(("token_type_hint", self.token.token_type_hint()));
+
+        Ok(endpoint_request(
+            self.auth_type,
+            self.client_id,
+            self.client_secret,
+            &self.extra_params,
+            None,
+            None,
+            self.revocation_url
```
ramosbugs, line 1689: I noticed that the [RFC](https://tools.ietf.org/html/rfc7009#section-2) says `Clients MUST verify that the URL is an HTTPS URL`. we should probably fail here for other schemes

src/revocation.rs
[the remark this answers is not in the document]
ramosbugs: that's probably all true in the vast majority of cases, but the motivation for having separate traits is to support non-standard use cases, so I think it's worth dropping the requirement

src/lib.rs
ramosbugs: I think there's been some related discussion in the past. if we put the type arguments on the function instead, that makes the call sites more complex since users will need to add type annotations there. the tradeoff here is that we have a more complex `Client` type definition with backward compatibility issues, but once you define a type alias for `Client`, the call sites become simpler

ximon18: Understood.

ximon18:
I've applied some of the changes, I'll do the rest as soon as I can but that may be a day or two.

src/basic.rs
```diff
@@ -81,6 +90,11 @@ pub type BasicTokenResponse = StandardTokenResponse<EmptyExtraTokenFields, Basic
 pub type BasicTokenIntrospectionResponse =
     StandardTokenIntrospectionResponse<EmptyExtraTokenFields, BasicTokenType>;
 
+///
+/// Basic OAuth2 revocable token type
+///
+pub type BasicRevocableToken = StandardRevocableToken;
```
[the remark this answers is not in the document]
ximon18: I agree with your reasoning, but note that one consequence of removing it will be that the definition of BasicClient will no longer consistently use BasicXxx types, it would then be:

```
pub type BasicClient = Client<
    StandardRevocableToken,
    BasicErrorResponse,
    BasicTokenResponse,
    BasicTokenType,
    BasicTokenIntrospectionResponse,
    BasicRevocationErrorResponse,
>;
```

Is that okay? It looks a bit strange, perhaps confusing?

src/lib.rs
ximon18: >now that we have two ErrorResponse type parameters, it's probably time to document them in the doc comment. these were probably self-evident before, but less so now

I pushed something for this, but I'm not convinced this is really what you had in mind.

ximon18:
With the last commit that added `pub` to `use RevocableToken` I am able to build a modified openidconnect-rs v1.0.1 to use the new revocation functionality (with a number of tweaks required in the oidc crate of course).

src/basic.rs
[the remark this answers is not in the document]
ramosbugs: looks ok to me! I agree consistency is important, but I think having separate `BasicRevocableToken` and `StandardRevocableToken` types that are interchangeable can also lead to confusion since people can wonder "what's the difference between these? when should I use one over the other?"

🤷‍♂️ seems like both solutions have potential downsides

src/lib.rs
ramosbugs: thanks for expanding the docs! I think they address my concerns around `TE`/`TRE`

src/tests.rs
```diff
@@ -1646,6 +1670,176 @@ fn test_token_introspection_successful_with_basic_auth_full_response() {
...
+            Some("https://revocation/url".parse().unwrap()),
+            HttpResponse {
+                status_code: StatusCode::OK,
+                headers: vec![(
+                    CONTENT_TYPE,
+                    HeaderValue::from_str("application/json").unwrap(),
+                )]
+                .into_iter()
+                .collect(),
+                body: b"{}".to_vec(),
```
ramosbugs, line 1771: let's change these response bodies to be empty now that the code handles that correctly

ramosbugs (commented):
looks like just the one test comment and then it'll be ready to merge :-)

src/tests.rs
ximon18: Ah, I left it like that because that’s what I saw being returned by Google. Maybe test with both with an actual empty body and empty JSON, and perhaps even with a different response content type?

ramosbugs: sounds good

ximon18: Done. Perhaps it would be nicer to have less repetition in the tests but verbose tests are obvious and truly isolated and independent of one another. Let me know if you think a helper method or something should be introduced to reduce the (near) repetition in these tests.

ramosbugs:
thanks! I'll cut a release once CI passes

ramosbugs:
https://crates.io/crates/oauth2/4.0.0-alpha.4

PR #123, opened 2021-02-13
Replace occurences of introspect with introspection. Improved some introspection related comments.

ximon18:
Be consistent with RFC-7662 and RFC-8414 which both refer to an introspection rather than introspect.


ramosbugs:
thanks!

PR #126, opened 2021-02-17
Add extra_fields() getter to StandardTokenIntrospectionResponse

seijikun:
The `StandardTokenIntrospectionResponse` struct was missing the `extra_fields()` getter, so I simply copied it from `StandardTokenResponse`.


ramosbugs:
thanks!

PR #127, opened 2021-02-18
Fail fast if endpoint URLs required by non-core endpoint requests are unusable

ximon18:
Fail fast if endpoint URLs required by non-core endpoint requests are missing or otherwise do not meet requirements.

Introduces a new `ConfigurationError` type for such configuration errors and modifies `fn exchange_device_code()`, `fn introspect()` and `fn revoke_token()` to use it.

Updates tests and examples accordingly.

Triggered by https://github.com/ramosbugs/openidconnect-rs/pull/39.

### DIFF
--- a/src/revocation.rs
+++ b/src/revocation.rs
@@ -1,4 +1,4 @@
-use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use serde::{Deserialize, Serialize};
 use std::fmt::Error as FormatterError;
 use std::fmt::{Debug, Display, Formatter};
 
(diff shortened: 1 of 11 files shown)


ramosbugs:
awesome, thanks!

src/revocation.rs
```diff
@@ -20,7 +20,7 @@ pub trait RevocableToken {
     /// Indicates the type of the token being revoked, as defined by [RFC 7009, Section 2.1](https://tools.ietf.org/html/rfc7009#section-2.1).
     ///
     /// Implementations should return `Some(...)` values for token types that the target authorization servers are
-    /// expected to know (e.g. because they are registered in the [OAuth Token Type Hints Registry](https://tools.ietf.org/html/rfc7009#section-4.1.2))
+    /// expected to know (e.g. because they are registered in the [OAuth Token Type Hints Registry](https://tools.ietf.org/html/rfc7009#section-4.1.2)
```
ramosbugs, line 23: I'm going to revert this one change since the trailing parenthesis corresponds to the one in the text preceding `e.g.`. that one is unterminated in the latest docs: https://docs.rs/oauth2/4.0.0-alpha.4/oauth2/revocation/trait.RevocableToken.html

ramosbugs: looks like you had already done that in #125 and then reverted your fix 😆

PR #128, opened 2021-02-23
Changes discussed on ramosbugs/openidconnect-rs#39.

ximon18:
See https://github.com/ramosbugs/openidconnect-rs/pull/39.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -870,6 +870,12 @@ where
     /// Query the authorization server [`RFC 7662 compatible`](https://tools.ietf.org/html/rfc7662) introspection
     /// endpoint to determine the set of metadata for a previously received token.
     ///
+    /// Requires that [`set_introspection_uri()`](Self::set_introspection_uri()) have already been called to set the
+    /// introspection endpoint URL.
+    ///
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -564,17 +564,11 @@ pub enum AuthType {
...
-///
-/// assert!(matches!(revocation_response, Err(
-///     RequestTokenError::ServerResponse(
-///         BasicRevocationErrorResponse{
-///             error: RevocationErrorResponseType::UnsupportedTokenType,
-///             ..
-///         })
-///     )
-/// ));
+///     .request(...);
```
ramosbugs, line 567: since `revoke_token` returns a `Result` now, the example would need `unwrap`/`expect` before calling `request`. might be worth replacing `ignore` with `rust` and trying to get the example to compile in order to prevent future inconsistencies

ximon18: Will do.

ximon18: Done.

ramosbugs:
thanks!

ramosbugs:
for posterity, I merged a follow-up to this to fix some inconsistencies: 4d55c26ad6e233d8f23b4514fe743d365a5a432f

PR #129, opened 2021-03-05
Allow to select reqwest TLS provider

co42:
Add two Cargo features so we can choose which TLS backend is used by reqwest.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,15 +15,17 @@ all-features = true
 maintenance = { status = "actively-developed" }
 
 [features]
-default = ["reqwest"]
+default = ["reqwest", "rustls-tls"]
 pkce-plain = []
+native-tls = ["reqwest/native-tls"]
+rustls-tls = ["reqwest/rustls-tls"]
 
 [dependencies]
 base64 = "0.12"
 thiserror="1.0"
 http = "0.2"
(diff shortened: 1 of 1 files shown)


ramosbugs:
thanks for the PR!

I'd like to figure out a way to accommodate this request, but I don't think we should introduce a breaking change to the feature flags to do so. In particular, the `reqwest` feature flag should continue to enable the `reqwest` client with rustls, which I think avoids an OpenSSL dependency for many users. This follows the broader direction of the Rust ecosystem in minimizing unsafe dependencies.

If you can figure out a clean way to add another feature flag while leaving the behavior of the `reqwest` flag unchanged, I'd be interested in merging it.

It's also worth pointing out that anyone can use `reqwest` with `native-tls` (or any other HTTP client) by providing their own `HttpRequest -> HttpResponse` implementation as described [here](https://docs.rs/oauth2/4.0.0-alpha.6/oauth2/#importing-oauth2-selecting-an-http-client-interface) under "Custom." It takes very little code to do that: https://github.com/ramosbugs/oauth2-rs/blob/main/src/reqwest.rs (especially since most users need either the sync or async implementation, but usually not both together)

ramosbugs:
thanks! I think this should work

PR #133, opened 2021-03-29
Errored->Failed

ximon18:
Partial revert of search-replace errors introduced in commit 09529db9bdc5059b7d0960b982486d42ea85f326.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3021,12 +3021,12 @@ where
     ServerResponse(T),
     ///
     /// An error occurred while sending the request or receiving the response (e.g., network
-    /// connectivity Errored).
+    /// connectivity failed).
     ///
-    #[error("Request Errored")]
+    #[error("Request failed")]
     Request(#[source] RE),
     ///
-    /// Errored to parse server response. Parse errors may occur while parsing either successful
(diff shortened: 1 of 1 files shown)


ximon18:
Hmm, the Travis buld seems to have failed somewhere unrelated to this change, some issue with a dependency. Does Travis use the `Cargo.lock` file? I wonder if it pulled in a newer version of a dependency that is breaking for some reason:

```
error[E0658]: `match` is not allowed in a `const fn`
   --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/socket2-0.4.0/src/lib.rs:156:9
    |
156 | /         match address {
157 | |             SocketAddr::V4(_) => Domain::IPV4,
158 | |             SocketAddr::V6(_) => Domain::IPV6,
159 | |         }
    | |_________^
    |
    = note: see issue #49146 <https://github.com/rust-lang/rust/issues/49146> for more information
error: aborting due to previous error
For more information about this error, try `rustc --explain E0658`.
error: could not compile `socket2`.
```

ramosbugs:
thanks! I'll look into the CI failures. there's been one issue involving 1.45, but now there's a new issue involving syn/js-sys. I wish people would stop breaking semver in critical parts of the ecosystem.

PR #138, opened 2021-04-21
Add add_scopes convenience method for many scopes and/or runtime defined

jeffrade:
This PR allows clients to pass a `&Vec<String>` to add desired `Scope`'s via new `add_scopes` method. When a client has more than one scope to add and knows them at build time, then the current `add_scope` suffices. But if the desired scopes are known at runtime (e.g. Users might click desired scopes from a list of UI check boxes), this method will be useful and less code for them to write.

No new `clippy` warnings and ran `cargo fmt`. Updated existing test to just use a single scope and added new test to use new `add_scopes` method so it is covered.

Was going to create an issue to discuss, but this was small enough to simply open a PR and discussion can happen here.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -991,6 +991,16 @@ impl<'a> AuthorizationRequest<'a> {
     }
 
     ///
+    /// Creates a new Scope for each Vector String and appends to the authorization URL.
+    ///
+    pub fn add_scopes(mut self, scopes: &[String]) -> Self {
+        scopes
+            .iter()
+            .for_each(|s| self.scopes.push(Cow::Owned(Scope::new(s.to_string()))));
+        self
+    }
+
+    ///
     /// Appends an extra param to the authorization URL.
     ///
     /// This method allows extensions to be used without direct support from
@@ -1318,6 +1328,16 @@ where
     }
 
     ///
+    /// Creates a new Scope for each Vector String and appends to the authorization URL.
+    ///
+    pub fn add_scopes(mut self, scopes: &[String]) -> Self {
+        scopes
+            .iter()
+            .for_each(|s| self.scopes.push(Cow::Owned(Scope::new(s.to_string()))));
+        self
+    }
+
+    ///
     /// Synchronously sends the request to the authorization server and awaits a response.
     ///
     pub fn request<F, RE>(self, http_client: F) -> Result<TR, RequestTokenError<RE, TE>>
@@ -1431,6 +1451,16 @@ where
     }
 
     ///
+    /// Creates a new Scope for each Vector String and appends to the authorization URL.
+    ///
+    pub fn add_scopes(mut self, scopes: &[String]) -> Self {
+        scopes
+            .iter()
+            .for_each(|s| self.scopes.push(Cow::Owned(Scope::new(s.to_string()))));
+        self
+    }
+
+    ///
     /// Synchronously sends the request to the authorization server and awaits a response.
     ///
     pub fn request<F, RE>(self, http_client: F) -> Result<TR, RequestTokenError<RE, TE>>
@@ -1540,6 +1570,16 @@ where
     ///
     pub fn add_scope(mut self, scope: Scope) -> Self {
         self.scopes.push(Cow::Owned(scope));
+        self
+    }
+
+    ///
+    /// Creates a new Scope for each Vector String and appends to the authorization URL.
+    ///
+    pub fn add_scopes(mut self, scopes: &[String]) -> Self {
+        scopes
+            .iter()
+            .for_each(|s| self.scopes.push(Cow::Owned(Scope::new(s.to_string()))));
         self
     }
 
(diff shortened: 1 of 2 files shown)


ramosbugs:
thanks for the PR, and apologies for the delayed response!

I think this use case (scopes determined dynamically at runtime) makes perfect sense, but I'm not sure `add_scopes` is substantially more convenient than having the caller use the existing `add_scope` method inside of a `for_each` call like the new `add_scopes` implementations are doing:
```rust
scopes
    .iter()
    .for_each(|s| request.add_scope(Scope::new(s.to_string())));
```

did you run into any trouble doing that that justified needing new methods on each of the request types?

as far as the code change itself, two minor comments:
1. for consistency with the strongly-typed interfaces, I'd suggest passing in `&[Scope]` instead of `&[String]`, though making the function generic and accepting `IntoIterator<Item = Scope>` would probably be best
2. instead of calling `Vec::push` repeatedly, we can call `Vec::extend` once, which gives it an opportunity to allocate any additional memory all at once instead of potentially having to grow and copy the contents multiple times if there are many scopes being added:
```rust
pub fn add_scopes<I>(mut self, scopes: I) -> Self
where I: IntoIterator<Item = Scope> {
    self.scopes.extend(scopes);
    self
}
```

thoughts?

jeffrade:
> did you run into any trouble doing that that justified needing new methods on each of the request types?

Hi @ramosbugs Yes, from the example in the docs:
```
let (auth_url, csrf_token) = client
  .authorize_url(CsrfToken::new_random)
  // Set the desired scopes.
  .add_scope(Scope::new("read".to_string()))
  .add_scope(Scope::new("write".to_string()))
  // Set the PKCE code challenge.
  .set_pkce_challenge(pkce_challenge)
  .url();
```
That works when I know that `read` and `write` are the desired scopes at compile time. But if I have one or more scopes to add based on runtime input, this PR just adds a new convenient `add_scopes` method to keep things concise and cleaner. I do see in your example I can simply iterate with the `request` and `add_scope` as needed but where would I do that given the example code above?

> as far as the code change itself, two minor comments:

I agree with both and can make those changes, but ran into a compiler error I tried resolving a few different ways without success.
```
error[E0271]: type mismatch resolving `<I as IntoIterator>::Item == Cow<'a, Scope>`
    --> src/lib.rs:1572:21
     |
1572 |         self.scopes.extend(scopes);
     |                     ^^^^^^ expected struct `Scope`, found enum `Cow`
     |
     = note: expected struct `Scope`
                  found enum `Cow<'a, Scope, >`
```
This is beyond my current knowledge of Rust (advanced Traits, etc.) so will have to play around with it some more. Will update here when resolved or come up with another solution that suffices "memory allocation once".

Thanks for reviewing!

ramosbugs:
> That works when I know that read and write are the desired scopes at compile time. But if I have one or more scopes to add based on runtime input, this PR just adds a new convenient add_scopes method to keep things concise and cleaner. I do see in your example I can simply iterate with the request and add_scope as needed but where would I do that given the example code above?

Each of the request types already has an `add_scope` method for adding one scope at a time, e.g. https://docs.rs/oauth2/4.0.0/oauth2/struct.AuthorizationRequest.html#method.add_scope. The code snippet above would be in your client code that uses this crate. So, I think this is already possible, but with potentially suboptimal performance when adding *many* scopes (though I doubt most use cases would be adding enough scopes for that to be noticeable).

> ran into a compiler error I tried resolving a few different ways without success.

ah whoops, forgot the `Vec` was of `Cow`s. this should hopefully resolve the compiler error:
```rust
pub fn add_scopes<I>(mut self, scopes: I) -> Self
where I: IntoIterator<Item = Scope> {
    self.scopes.extend(scopes.into_iter().map(Cow::Owned));
    self
}
```

in any case, it sounds like the motivation for this PR may have been not realizing that there were already `add_scope` methods on each of the request types, and the additional usefulness of `add_scopes` methods may be limited.

jeffrade:
> ah whoops, forgot the Vec was of Cows. this should hopefully resolve the compiler error:

Thank you @ramosbugs now fixed.

Updated our code based on your suggestion and now it looks like this:
```
    let mut auth_request = client
        .authorize_url(|| CsrfToken::new(csrf_token_string))
        .set_pkce_challenge(pkce_challenge);
    for s in scopes.iter() {
        auth_request = auth_request.add_scope(Scope::new(s.to_string()));
    }
    let (auth_url, csrf_token) = auth_request.url();
```

IMHO it would be cleaner and more convenient to just simply do:
```
    // scopes is of type Vec<Scope>
    let (auth_url, csrf_token) = client
        .authorize_url(|| CsrfToken::new(csrf_token_string))
        .add_scope(scopes)
        .set_pkce_challenge(pkce_challenge);
        .url();
```

But no worries if you prefer not to merge since there's no real fix here. Thanks again for reviewing!

ramosbugs:
Looks good now, thank you!

PR #143, opened 2021-05-22
Update Microsoft example to use unified Azure app registration

jongiddy:
The original method of registering an app to run this example no longer works. Update the comments to provide the new method.

See https://developer.microsoft.com/en-us/identity/blogs/deprecation-of-app-registrations-legacy-experience-and-the-application-registration-portal-apps-dev-microsoft-com/


ramosbugs:
Thanks!

PR #144, opened 2021-05-30
Add ability to override redirect_url on code exchange.

smurfpandey:
Adds method and test to override redirect_url on code exchange. With #95 we can override redirect url during authorization, this change adds similar functionality during code exchange.


ramosbugs:
Thanks for the PR, and for adding a test!

PR #145, opened 2021-05-31
Update dependencies (base64, rand, hmac)

Cogitri:


Cogitri:
Ah, seems like we need to adjust some things for the WASM targets

ramosbugs:
Thanks for the PR! https://rust-random.github.io/book/update-0.8.html#dependencies looks like it has some info about WASM targets and `getrandom`.

Cogitri:
Thanks for the link, should build now :)

ramosbugs:
Thank you!

PR #147, opened 2021-06-07
fix curl client implementation

maurerdietmar:
Note: write_function() callback is used to read data in GET request.


ramosbugs:
thanks for the PR! looks like this has been broken for ~2 years 😱

PR #149, opened 2021-06-29
Don't depend on chrono oldtime

Cogitri:
We don't need the time::Duration type, so we can get rid of the
(deprecated) time 0.1 dependency by disabling the oldtime chrono
feature


ramosbugs:
thanks! I just noticed Travis CI builds have apparently stopped since I ran out of open source credits under their new pricing model... sigh, I'll migrate this repo over to GitHub Actions soon!

PR #152, opened 2021-07-05
Await the sleep_fn future in DeviceAccessTokenRequest::request_async

Joey9801:
Previously the future returned by the sleep_fn wasn't awaited, such that DeviceAccessTokenRequest::request_async would spam requests to the token endpoint in a tight loop.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2326,7 +2326,7 @@ where
             }
 
             // Sleep here using the provided sleep function.
-            sleep_fn(interval);
+            sleep_fn(interval).await;
         }
     }
 


ramosbugs:
nice catch, and thanks for the fix! somewhat surprising that clippy doesn't warn on the unused future...

Joey9801:
It's a little surprisng yeah. Perhaps it could be related to how the effect of `#[must_use]` doesn't propagate through `Fn` objects either. Eg:
```rust
#[must_use]
fn foo() -> i32 {
    1337
}

fn bar(f: impl Fn () -> i32) {
    f();       // <-- Doesn't warn
}

fn main() {
    foo();     // <-- Does warn 
    bar(foo);  // <-- Doesn't warn
}
```

In any case, thanks for creating+maintaining this crate! 😄

PR #154, opened 2021-08-24
Fix typo in documentation

sindreij:
I got a bit confused that the docs mentioned reqwest 0.10, fixed it to the correct 0.11


ramosbugs:
thanks for the PR, and sorry about the stale docs!

PR #166, opened 2022-01-20
Make Client struct's client_id field to pub

### LINKED ISSUE #165: Could you make Client's client id as pub?
Since Twitter's OAuth2 requires the client ID in the parameter as well, even though the urlencoded client ID is already included in the request body ( https://github.com/ramosbugs/oauth2-rs/blob/main/src/lib.rs#L1935 ), I would like Client struct to exposes the client_id parameter as pub or adding pub function.

I made little workaround to use this great library on twitter oauth2 as like
```Rust
// Get an auth token
let token = oauth_client
    .exchange_code(code)
    .add_extra_param("client_id", env::var("TWITTER_CLIENT_ID").context("Missing TWITTER_CLIENT_ID!").unwrap()) // Here!
    .set_pkce_verifier(session.get("pkce_code_verifier").unwrap())
    .request_async(async_http_client)
    .await
    .unwrap();
```

tomocrafter:
closes #165

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -616,7 +616,7 @@ where
     RT: RevocableToken,
     TRE: ErrorResponse,
 {
-    client_id: ClientId,
+    pub client_id: ClientId,
     client_secret: Option<ClientSecret>,
     auth_url: AuthUrl,
     auth_type: AuthType,


tomocrafter:
Oh wait, He said all fields not only for client_id.
https://github.com/ramosbugs/oauth2-rs/issues/165#issuecomment-1017071198

I think I should make new PR that makes all fields to pub so I close this PR for now.

PR #167, opened 2022-01-20
Make pub all fields in Client struct

### LINKED ISSUE #165: Could you make Client's client id as pub?

tomocrafter:
closes #165


ramosbugs:
thanks for the PR! mind adding getter functions rather than directly exposing the fields? I think the main difference is that making the fields `pub` would allow for mutation, but I think an immutable interface is what we want here.

e.g.
```rust
pub fn client_id(&self) -> &ClientId {
  &self.client_id
}
```

tomocrafter:
Should I put the details of the return value of the function in the doc comment of each function?

src/lib.rs
```diff
@@ -965,6 +965,42 @@ where
             _phantom: PhantomData,
         })
     }
+
+    pub fn client_id(&self) -> &ClientId {
+        &self.client_id
+    }
+
+    pub fn client_secret(&self) -> Option<&ClientSecret> {
```
ramosbugs, line 973: let's leave out the client secret since it's safer for secrets to be write-only

ramosbugs (commented):
> Should I put the details of the return value of the function in the doc comment of each function?

I don't have a strong opinion, but clippy is sad at the moment and it's probably better to put a short doc comment rather than disable the lint. Feel free to copy/paste the description from the setter and/or factory method

thanks for taking care of this!

PR #171, opened 2022-02-18
breaking change: use a helper type to accept static `str`s as well as heap-allocated `String`s in builders and API

WilliamVenner:
I have refactored parts of the API to accept `str` as well as `String` to reduce unnecessary allocations.

I decided to not use `Cow<'static, str>` and instead made the `AsStr` type in order to implement the `as_str` function for backwards compatibility. Despite backwards-compatibility-aware changes, this may introduce breaking changes for some users, so a major version change would be required for this PR.


ramosbugs:
See discussion in #248

PR #174, opened 2022-04-29
Update dependencies

jelmansouri:
We use cargo deny to lint our dependencies on my project, sha2 seems is out of date, but I took the opportunity to update the other dev deps. Hopefully this is all still compliant with the MSRV.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ curl = { version = "0.4.0", optional = true }
 [dev-dependencies]
 hex = "0.4"
 hmac = "0.12"
-uuid = { version = "1.0", features = ["v4"] }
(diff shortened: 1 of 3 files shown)


ramosbugs:
Thanks for the PR! Looks like the `uuid` bump broke MSRV though 😞

jelmansouri:
Yeah it seems like the MSRV for uuid is 1.46 since they rely on some functions made const. I'll bump back uuid!

ramosbugs:
Thanks!

ramosbugs:
This is now released in https://crates.io/crates/oauth2/4.2.0

jelmansouri:
You rock ! Thanks a lot !

PR #180, opened 2022-08-06
(breaking): Reduce unnecessary lifetimes

### LINKED ISSUE #179: Implicitly captured `self` on return Futures
When using `request_async`, the `async` function implicity captures `self` and therefore can lead to restricted lifetimes. For example

Given the following wrapper around `BasicClient`

```rust
pub struct Oauth2Client(oauth2::basic::BasicClient);

impl Oauth2Client {
    pub fn request_token(
        &self,
        code: AuthorizationCode,
        verifier: PkceCodeVerifier,
    ) -> impl Future<
        Output = Result<
            BasicTokenResponse,
            RequestTokenError<AsyncHttpClientError, BasicErrorResponse>,
        >,
    > {
        self.0
            .exchange_code(code)
            .set_pkce_verifier(verifier)
            .request_async(oauth2::reqwest::async_http_client)
    }
}
```

This won't compile with the following error:

```
error[E0700]: hidden type for `impl Trait` captures lifetime that does not appear in bounds
  --> src/main.rs:17:9
   |
8  |           &self,
   |           ----- hidden type `impl Future<Output = Result<StandardTokenResponse<EmptyExtraTokenFields, BasicTokenType>, oauth2::RequestTokenError<oauth2::reqwest::Error<reqwest::error::Error>, StandardErrorResponse<BasicErrorResponseType>>>>` captures the anonymous lifetime defined here
...
17 | /         self.0
18 | |             .exchange_code(code)
19 | |             .set_pkce_verifier(verifier)
20 | |             .request_async(oauth2::reqwest::async_http_client)
   | |______________________________________________________________^
   |
help: to declare that the `impl Trait` captures `'_`, you can add an explicit `'_` lifetime bound
   |
16 |     > + '_ {
```

This is because `CodeTokenRequest` has a lifetime associated with `Client` and then subsequently gets captured in the future generated by `request_async`. We can modify `request_async` to return a non-capturing future instead.

```rust
pub fn request_async<C, F, RE>(
      self,
      http_client: C,
  ) -> impl Future<Output = Result<TR, RequestTokenError<RE, TE>>>
  where
      C: FnOnce(HttpRequest) -> F,
      F: Future<Output = Result<HttpResponse, RE>>,
      RE: Error + 'static,
  {
      futures::future::ready(self.prepare_request()).and_then(|http_request| {
          http_client(http_request).map(|res| {
              res.map_err(RequestTokenError::Request)
                  .and_then(endpoint_response)
          })
      })
  }
```

which will then not have the implicit lifetime and allow for the previous code to compile. I'm happy to implement this. I've used `futures` here which isn't a dependency, but I could create a custom Future as all `request_async` functions are basically the same with the exception `DeviceAccessTokenRequest::request_async`.

jpopesculian:
Fixes #179 

This PR attempts to reduce unnecessary lifetimes. This is unfortunately at the expense of a little bit more complexity in handling the Futures. This effect could maybe be reduced somewhat with the use of the `futures` crate.

The other significant change here is the change from `Arc<Fn() -> DateTime + 'b>` to a type parameter which reduced the complexity of the future and I believe reduced the complexity of the lifetimes of `DeviceAcessTokenRequest`. However this is a breaking change!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ ureq = { version = "2", optional = true }
 url = { version = "2.1", features = ["serde"] }
 chrono = { version = "0.4", default-features = false, features = ["clock", "serde", "std"] }
 serde_path_to_error = "0.1"
+pin-project = "1.0"
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 getrandom = { version = "0.2", features = ["js"] }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -425,11 +425,14 @@ use std::fmt::Error as FormatterError;
 use std::fmt::{Debug, Display, Formatter};
 use std::future::Future;
 use std::marker::PhantomData;
-use std::sync::Arc;
+use std::mem::MaybeUninit;
+use std::pin::Pin;
+use std::task::{Context, Poll};
 use std::time::Duration;
 
 use http::header::{HeaderMap, HeaderValue, ACCEPT, AUTHORIZATION, CONTENT_TYPE};
 use http::status::StatusCode;
+use pin_project::pin_project;
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 use url::{form_urlencoded, Url};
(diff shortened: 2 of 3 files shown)


ramosbugs:
See comment on #179

Issue #182, opened 2022-09-08
ureq `http_client` doesn't work with microsoft (tested for devicecode flow)


nerditation:
I was trying to authenticate using microsoft devicecode flow, as documented [here](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-device-code), but the server returned 404. but I got a totally normal response when I directly post the authentication data to the endpoint using `ureq::Request::send_form(...)`

so I checked the logs and found that there were different http headers: the failed on has a `Transfer-Encoding: chunked` header, while the good one contains a `Content-Length: 12345678` header.

I don't know whether microsoft's implementation is correct, or what the oauth standard says, but I found the reason leading to this difference is `oauth2::ureq::http_client` calls `send(reader)` on the `ureq::Request` object:
https://github.com/ramosbugs/oauth2-rs/blob/93452e57b52a9ecd374e7b4293836465d70caad6/src/ureq.rs#L55
but we could instead (and should) call `send_bytes(byte_slice)` since the body is already serialized data (urlencoded form), why add another indirection to wrap it into a reader?
```rust
    req.send_bytes(&request.body)
```
this not only feels more nature, it also fixes the compatibility issue with microsoft since payload with type of `slice` makes ureq automatically add a `Content-Length` header.

ramosbugs:
Sounds like a good fix. Please feel free to submit a PR!

PR #183, opened 2022-09-09
ureq::http_client use `send_bytes()` for POST requests, solves #182

nerditation:
`ureq::Request::send_bytes` takes a byte slice as argument, so ureq knows the size of the payload and would insert a `Content-Length` http header, while `ureq::Request::send` takes a `impl Read` that streams the payload data, thus the `Content-Length` cannot be  calculated beforehand.


ramosbugs:
Thanks!

PR #186, opened 2022-09-25
feat: add const context to wrapper types

MrNossiom:


ramosbugs:
Seems like a worthwhile improvement, thanks!

MrNossiom:
Thanks. I don't know if I should also add it to the secret wrapper types, since there is a small chance anyone would hardcode their tokens. Not doing so may prevent bad practice.

ramosbugs:
Agreed, let's just do the non-secret types.

MrNossiom:
I think this is the only place adding `const` is useful.
Thanks

PR #192, opened 2022-11-03
Implement more From and TryFrom traits

henrikhorluck:
- `From` over `Into` for all new_type, since you then get `Into` for free
- Add `From` to secret types, since it only allows transforms to the
  ecret type, not from it
- Add `FromStr` and `TryFrom` to new_url_type for better ergonomics

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -1,6 +1,7 @@
 use http::header::{HeaderMap, HeaderName, HeaderValue, ACCEPT, AUTHORIZATION, CONTENT_TYPE};
 use http::status::StatusCode;
 use revocation::RevocationErrorResponseType;
+use std::convert::TryInto;
 use thiserror::Error;
 use url::form_urlencoded::byte_serialize;
 use url::Url;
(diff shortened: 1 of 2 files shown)


ramosbugs:
Hey, thanks for the PR, but the lack of `From`/`TryFrom` is an intentional design decision to prevent accidental coercions into each of the "new types." The `new()` method is the sole way (other than implicitly via deserialization) to construct each of these types, serving as an explicit assertion in the code that the value is of the given type (e.g., `ClientId::new(client_id_str)` vs. `client_id_str.into()`).

Otherwise, it's entirely too easy to accidentally coerce a string into the wrong new type when passing arguments to a function (e.g., maybe `client_id_str` gets coerced into a `ClientSecret`, which would not be obvious from the code calling `client_id_str.into()`). Being able to prevent that sort of bug via the type system is one of the key benefits of Rust.

henrikhorluck:
>  e.g., maybe client_id_str gets coerced into a ClientSecret, which would not be obvious from the code calling client_id_str.into()

But the way you have set it up, it seems obvious that any sensible programmer would instead set it up correctly way earlier with the explicit types, e.g. have the following as soon as possible:

```rust
// assume client_id is a String
let client_id: ClientId = client_id.into();
```

And with e.g. clap's argument parsing, you can even completely avoid ever having an `&str` or `String` in your code 
([example](https://github.com/henrikhorluck/emne-eksport/blob/main/src/main.rs#L121)), which the lack of an `From<String>` makes much less ergonomic than necessary.

Yes, Rust is a great language for safety thanks to it's type system, but it is verbose and very syntax-heavy, and the lack of these `impl`s just further worsens it. From/into is nowhere near as confusing as Deref, yet you implement it, going against the two API guideline  [Only smart pointers implement Deref and DerefMut](https://rust-lang.github.io/api-guidelines/predictability.html#only-smart-pointers-implement-deref-and-derefmut-c-deref) and [Conversions use the standard traits From, AsRef, AsMut](https://rust-lang.github.io/api-guidelines/interoperability.html#conversions-use-the-standard-traits-from-asref-asmut-c-conv-traits)

PR #195, opened 2022-11-07
Use DeviceCodeErrorResponseType::ExpiredToken for expired Device

LorenzoLeonardo:
Hello David

Here is the changed to use the DeviceCodeErrorResponseType::ExpiredToken and the Unit Test changes.

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -2203,7 +2203,14 @@ fn test_device_token_authorization_timeout() {
         .err()
         .unwrap();
     match token {
-        RequestTokenError::Other(msg) => assert_eq!(msg, "Device code expired"),
+        RequestTokenError::ServerResponse(msg) => assert_eq!(
+            msg,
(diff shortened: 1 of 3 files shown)


src/tests.rs
```diff
@@ -2203,6 +2203,14 @@ fn test_device_token_authorization_timeout() {
...
     match token {
+        RequestTokenError::ServerResponse(msg) => assert_eq!(
+            msg,
+            DeviceCodeErrorResponse::new(
+                DeviceCodeErrorResponseType::ExpiredToken,
+                Some(String::from("This device code has expired.")),
+                None,
+            )
+        ),
         RequestTokenError::Other(msg) => assert_eq!(msg, "Device code expired"),
```
ramosbugs, line 2214: let's remove line 2214 since we no longer expect that response

ramosbugs (commented):
thanks for the PR!

PR #197, opened 2022-11-14
Fix serde_path_to_error dependency

cschramm:
It is expected that `serde_path_to_error::Error` implements `Debug` but that's only the case from 1.0.2 on.


ramosbugs:
Thanks!

PR #203, opened 2023-01-20
Add cancellation when polling of access token

LorenzoLeonardo:
There is a use case that a user application in the device code workflow that cancels during login phase

The polling of access token is immediately start running after the user code and the url was already given to the user.

The idea of this is that the user application can create their own http_client and be able to return an Error result when cancelling.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2432,7 +2432,16 @@ where
     {
         let http_response = match res {
             Ok(inner) => inner,
-            Err(_) => {
+            Err(err) => {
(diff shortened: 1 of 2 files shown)


LorenzoLeonardo:
Hello David 


I have another idea on how to cancel polling for an access token.
https://github.com/ramosbugs/oauth2-rs/pull/203. Let me know if we can improve this.

LorenzoLeonardo:
Not needed anymore, closing this issue

Issue #214, opened 2023-05-15
Compounding the sleep intervals if there is an error in the http_client


LorenzoLeonardo:
https://github.com/ramosbugs/oauth2-rs/blame/00fa30b336a8b3c9c4c900ae3ccba371459c0ac8/src/lib.rs#L2435

It was observed that the sleep interval if there is an error in the http_client is compounding. Bad for user experience, it will wait them too long to get the token example if ethernet cable is unplugged for a long time, then plugged it again and login in their cloud account. How about just doubling it but not compounding?

LorenzoLeonardo:
Here is the pull request I made to fix the issue of compounding.
https://github.com/ramosbugs/oauth2-rs/pull/215

ramosbugs:
Hmm... the [relevant part of the spec](https://www.rfc-editor.org/rfc/rfc8628#section-3.5) here is:
>  On encountering a connection timeout, clients MUST unilaterally
   reduce their polling frequency before retrying.  The use of an
   exponential backoff algorithm to achieve this, such as doubling the
   polling interval on each such connection timeout, is RECOMMENDED.

A connection timeout seems similar to the `slow_down` case except where the server is so overloaded it can't process the request and return `slow_down`. For `slow_down`, the spec says:
>  A variant of "authorization_pending", the authorization request is
      still pending and polling should continue, but the interval MUST
      be increased by 5 seconds for this and all subsequent requests.

I think it's reasonable to set an upper bound on the polling interval so that it doesn't double to an effectively infinite backoff in the scenario you're describing, but some backoff is explicitly required by the spec. Unfortunately, the generic `http_client` interface's error type doesn't provide us a way to distinguish connection timeouts from other types of request errors. A disconnected ethernet cable would probably result in a harder error than a timeout since, e.g., DNS resolution would potentially fail.

I propose we leave the doubling logic but set a default upper bound to 10 seconds (twice the default 5-second interval mentioned throughout the spec). It's potentially also worth adding a `set_max_backoff_interval` method to `DeviceAccessTokenRequest` so that clients can override the 10-second default behavior.

For clarity, I'm proposing changing this:
https://github.com/ramosbugs/oauth2-rs/blob/00fa30b336a8b3c9c4c900ae3ccba371459c0ac8/src/lib.rs#L2436-L2437
to something like:
```rust
// RFC 8628 requires a backoff in cases of connection timeout, but we can't
// distinguish between connection timeouts and other HTTP client request errors
[...]

PR #216, opened 2023-05-16
Fixing sleep interval that compounds

LorenzoLeonardo:
It was observe that during polling of token, if an error keeps returning from the http_client, the sleep interval compounds by 2. This scenario can be reproduced if ethernet cable is unplugged for a long time not more than the expiry time.

To reproduced, set up a custom curl http_client with connect_timeout() and timeout() from Easy2 curl-rust crate.

Thus, a max back off interval is introduce to up a limit to the compounding of time interval.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -897,6 +897,7 @@ where
             token_url: self.token_url.as_ref(),
             dev_auth_resp: auth_response,
             time_fn: Arc::new(Utc::now),
+            max_backoff_interval: None,
             _phantom: PhantomData,
         }
     }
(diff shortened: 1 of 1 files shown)


LorenzoLeonardo:
@ramosbugs  solution for issue https://github.com/ramosbugs/oauth2-rs/issues/214

ramosbugs:
thanks!

LorenzoLeonardo:
Hello David

Do you have plans to version up the crate? I will use it with this fix.
Thanks a lot.

Best Regards

Lorenzo

ramosbugs:
[4.4.0](https://crates.io/crates/oauth2/4.4.0) is now published

LorenzoLeonardo:
Thank you very much.

Issue #219, opened 2023-06-14
microsoft_devicecode flow broken


randomairborne:
When attempting to run the microsoft_devicecode example with my azure AD tenant, I get 
```Error: ServerResponse(StandardErrorResponse { error: invalid_scope, error_description: Some("AADSTS70011: The provided value for the input parameter 'scope' is not valid. The scope 'read' does not exist.\r\nTrace ID: 7172ec82-5230-49ab-b98c-a224c8e50500\r\nCorrelation ID: afde4e78-9981-460a-a670-e3ef2679d6dd\r\nTimestamp: 2023-06-14 23:12:54Z"), error_uri: None })```
and when switching it to `XboxLive.signin`, after `exchange_device_access_token` i get `Error: Other("Server returned empty error response")`.

ramosbugs:
@LorenzoLeonardo Do you happen to remember if there's anything that needs to be configured on the Azure side in order for the example to work? Sounds like there might be some missing documentation, but I'm not familiar enough with Azure to know what steps are needed.

randomairborne:
Note: I also switched from `common` to `consumers` to match my tenant type

ramosbugs:
> Note: I also switched from `common` to `consumers` to match my tenant type

ah, that's probably why. Is this the same as Azure AD B2C? See https://learn.microsoft.com/en-us/azure/active-directory-b2c/access-tokens#scopes.

randomairborne:
switching it back didn't seem to fix anything, though?

randomairborne:
I will end up getting a device code and the link i need, but then it will immediately return `Error: Other("Server returned empty error response")` with no further details

ramosbugs:
This error means the server returned 200 OK but no response body, so I'm not sure what additional debug information this crate could provide in that situation. There might be some informative response headers, which you could access by providing a custom `http_client` function to `.request_async()` that wraps whichever HTTP client you're using and looks at the response before passing it back to the crate.

In any case, this isn't an Azure support forum, and it sounds like you're trying to use the example on a different type of tenant than the example was written for. Feel free to contribute a working B2C example if you get it working, but the exact input parameters needed to do that are specific to each provider and not something this crate aims to implement for every possible provider configuration.

LorenzoLeonardo:
[...]

PR #220, opened 2023-06-15
Fix token url for device code flow

randomairborne:
this should fix ramosbugs/oauth2-rs#219, as according to https://learn.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-device-code that was the wrong token url


ramosbugs:
@LorenzoLeonardo seems like the order of `common/oauth2` vs. `oauth2/common` within the URL may have been an issue?

randomairborne:
@ramosbugs that was in fact the issue

ramosbugs:
@LorenzoLeonardo I'm going to merge this unless you have objections

ramosbugs:
Thanks for fixing!

PR #221, opened 2023-06-15
Add Microsoft device code example with tenant id

LorenzoLeonardo:
To make it more understandable for Microsoft endpoint users,
an endpoint with tenant is added and edited the existing
example as common user.


LorenzoLeonardo:
Wrong PR

PR #222, opened 2023-06-15
Add Microsoft device code example with tenant id

LorenzoLeonardo:
To make it more understandable for Microsoft Azure endpoint users, an endpoint with tenant is added and edited the existing example as common user.


examples/microsoft_devicecode_tenant_user.rs
```diff
@@ -0,0 +1,44 @@
...
+        "https://login.microsoftonline.com/{tenant}/oauth2/v2.0/devicecode".to_string(),
+    )?;
+    let client = BasicClient::new(
+        ClientId::new("client_id".to_string()),
+        None,
+        AuthUrl::new(
+            "https://login.microsoftonline.com/{tenant}/oauth2/v2.0/authorize".to_string(),
+        )?,
+        Some(TokenUrl::new(
+            "https://login.microsoftonline.com/{tenant}/v2.0/oauth2/token".to_string(),
```
ramosbugs, line 19: shouldn't this be `{tenant}/oauth2/v2.0/token`?

examples/microsoft_devicecode_tenant_user.rs
```diff
@@ -0,0 +1,44 @@
+use oauth2::basic::BasicClient;
+use oauth2::devicecode::StandardDeviceAuthorizationResponse;
+use oauth2::reqwest::async_http_client;
+use oauth2::{AuthUrl, ClientId, DeviceAuthorizationUrl, Scope, TokenUrl};
+use std::error::Error;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+    let device_auth_url = DeviceAuthorizationUrl::new(
+        "https://login.microsoftonline.com/{tenant}/oauth2/v2.0/devicecode".to_string(),
```
ramosbugs, line 10: to make it clear to users that this needs to be substituted for their tenant, I'd suggest defining a `TENANT_ID` global at the top, and then using the `format!` macro here to automatically fill it in so that they only have to update it in one place at the top. and then also a comment next to that global instructing them to do so, maybe with a link for where they can find that tenant ID

examples/microsoft_devicecode_tenant_user.rs
LorenzoLeonardo: I will change it. Thanks for finding it :)

ramosbugs: this is the same issue @randomairborne fixed in #220 for the common case

examples/microsoft_devicecode_tenant_user.rs
LorenzoLeonardo: Done

examples/microsoft_devicecode_tenant_user.rs
LorenzoLeonardo: @ramosbugs  ohh, I already changed that in this PR.

LorenzoLeonardo: @ramosbugs  ok you can merge it, I will just do a rebase

LorenzoLeonardo: @ramosbugs  rebase done

ramosbugs:
Thanks!

PR #227, opened 2023-06-23
Upgrade `base64` to `0.21`

MarijnS95:
A new trait-based system has been employed, with implementations for all the configurations rather than those configurations being passed as argument to the free `encode_config()` function.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ native-tls = ["reqwest/native-tls"]
 rustls-tls = ["reqwest/rustls-tls"]
 
 [dependencies]
-base64 = "0.13"
+base64 = "0.21"
 thiserror = "1.0"
 http = "0.2"
 rand = "0.8"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -427,6 +427,7 @@ use std::marker::PhantomData;
 use std::sync::Arc;
 use std::time::Duration;
 
+use base64::prelude::*;
 use chrono::serde::ts_seconds_option;
 use chrono::{DateTime, Utc};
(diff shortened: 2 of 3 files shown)


ramosbugs:
This crate's MSRV goes back even further than `openidconnect` to Rust 1.45, which I guess is incompatible with this version of `base64`'s use of the 2021 edition.

Whenever version 5.0 of this crate happens (no immediate plans), I'll introduce the same 6-month relaxed MSRV policy that I did in the latest major release of `openidconnect`. But, that change in policy feels like a breaking change, so I'd prefer to wait unless there's a compelling reason (e.g., serious bug fix or security issue).

MarijnS95:
@ramosbugs thanks! No immediate bugs/security issues AFAIK (or RUSTSEC will ring some bells), just pruning duplicates in my dependency trees as most of the ecosystem seems to have adopted `base64 0.21` now.

CosminPerRam:
Hey, just checking in the progress around this crate, would the [db0ea44](https://github.com/ramosbugs/oauth2-rs/commit/db0ea44657bd6c1130b83ce135ac2691ba091fad) commit replace this PR as it's already upgraded in main now?

ramosbugs:
> Hey, just checking in the progress around this crate, would the [db0ea44](https://github.com/ramosbugs/oauth2-rs/commit/db0ea44657bd6c1130b83ce135ac2691ba091fad) commit replace this PR as it's already upgraded in main now?

Yes, both `base64` 0.21 and 0.22 are supported now as of #261. Thanks @MarijnS95 for the PR. I forgot about it and did the upgrade separately when I was doing the breaking `reqwest` version bump 😅

PR #231, opened 2023-09-08
feat: Derive hash and eq for state

kate-shine:
Deriving traits necessary for storing CsrfToken in HashMap / HashSet - like structures for easier implementation of state store without a need to access the wrapped secret directly.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -579,7 +579,7 @@ new_secret_type![
     /// via the `state` parameter.
     ///
     #[must_use]
-    #[derive(Clone, Deserialize, Serialize)]
+    #[derive(Clone, Deserialize, Serialize, Hash, PartialEq, Eq)]
     CsrfToken(String)
     impl {
         ///


ramosbugs:
Hey, thanks for the PR. However, these traits are omitted from this crate's secret types intentionally to prevent timing attacks. Please see #107 for a more complete explanation.

I'd encourage developers not to use `CsrfToken` or any other sensitive types as hash table keys. Instead, I would use a secure hash (e.g., SHA-256) of the sensitive value as the hash key, which is safe against timing attacks. I believe Rust's default hash function for HashSet, etc. is designed to be safe against DoS attacks caused by too many entries aliasing to the same hash, but I don't think it's intended to prevent timing attacks on secret keys.

kate-shine:
Oof, yep, issue of timing attacks is always easy to miss. When I have time, I will implement it as feature as suggested in linked issue, thanks :)

PR #232, opened 2023-09-11
implementing feature for PartialEq and Hash with comparison of Sha256 hashed values for secrets

kate-shine:
following advice in
https://github.com/ramosbugs/oauth2-rs/issues/107#issuecomment-649834938
added a feature that allows safer comparisons of secret values that should be resistant to timing attacks.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ default = ["reqwest", "rustls-tls"]
 pkce-plain = []
 native-tls = ["reqwest/native-tls"]
 rustls-tls = ["reqwest/rustls-tls"]
+timing-resistant-secret-traits = []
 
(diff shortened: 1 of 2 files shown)


ramosbugs:
thanks for the PR! I think `cargo fmt` is complaining about trailing whitespace

kate-shine:
> thanks for the PR! I think `cargo fmt` is complaining about trailing whitespace

Thanks, fixed! It might be a good idea to add a more opinionated `rustfmt.toml` to repo :) I didn't autoformate, because my rustfmt settings changed way more than what my contribution contained.

Thanks for great crates, by the way :) I'm currently using them for MS Entra and it's a super smooth experience.

PR #238, opened 2023-11-26
reqwest: Migrate to shared `Error` type and use `thiserror`'s `From` impl

MarijnS95:
`reqwest` no longer has a separate error type for sync and async implementations.  In addition `thiserror` can generate `From` implementations for automatic conversions in the `Try` (`?`) operator.

Finally, delete some unnecessary `#[cfg]` wraps that are already guarding the module as a whole, and that the code won't compile without anyway.

### DIFF
--- a/src/reqwest.rs
+++ b/src/reqwest.rs
@@ -3,39 +3,24 @@ use thiserror::Error;
 ///
 /// Error type returned by failed reqwest HTTP requests.
 ///
+#[non_exhaustive]
 #[derive(Debug, Error)]
-pub enum Error<T>
-where
-    T: std::error::Error + 'static,
-{
+pub enum Error {
     /// Error returned by reqwest crate.
     #[error("request failed")]
-    Reqwest(#[source] T),
+    Reqwest(#[from] reqwest::Error),
     /// Non-reqwest HTTP error.
     #[error("HTTP error")]
-    Http(#[source] http::Error),
+    Http(#[from] http::Error),
     /// I/O error.
     #[error("I/O error")]
-    Io(#[source] std::io::Error),
-    /// Other error.
-    #[error("Other error: {}", _0)]
-    Other(String),
+    Io(#[from] std::io::Error),
 }
 
 #[cfg(not(target_arch = "wasm32"))]
 pub use blocking::http_client;
-///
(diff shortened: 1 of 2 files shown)


src/reqwest.rs
```diff
@@ -4,39 +4,23 @@ use thiserror::Error;
...
+    Reqwest(#[from] reqwest::Error),
     /// Non-reqwest HTTP error.
     #[error("HTTP error")]
-    Http(#[source] http::Error),
+    Http(#[from] http::Error),
     /// I/O error.
     #[error("I/O error")]
-    Io(#[source] std::io::Error),
-    /// Other error.
-    #[error("Other error: {}", _0)]
```
ramosbugs, line 21: If we're going to remove this variant, we should probably make the enum `#[non_exhaustive]` so we'll still have an escape hatch to avoid future breaking changes. In any case, we'll need `Other` or another variant for errors that arise during the `url::Url` to `http::Uri` conversion.

src/reqwest.rs
```diff
@@ -51,36 +35,29 @@ mod blocking {
...
             .request(request.method, request.url.as_str())
             .body(request.body);
 
         for (name, value) in &request.headers {
             request_builder = request_builder.header(name.as_str(), value.as_bytes());
         }
-        let mut response = client
-            .execute(request_builder.build().map_err(Error::Reqwest)?)
-            .map_err(Error::Reqwest)?;
+        let mut response = client.execute(request_builder.build()?)?;
```
ramosbugs, line 51: nice, I should use `#[from]` more

ramosbugs (commented):
Thanks for the PR!

src/reqwest.rs
MarijnS95: Sounds good; let's go for `non_exhaustive` to have an escape-hatch until we define what the error type is going to be for that conversion.

src/reqwest.rs
MarijnS95: Thanks! I've got some preliminary changes lying around to use this pattern more inside this crate, will see to proposing them in a PR.

MarijnS95: Done in https://github.com/ramosbugs/oauth2-rs/pull/2391

PR #239, opened 2023-11-28
Replace `map_err()` conversions with a `From` call via the `Try` operator

MarijnS95:
The `?` or `Try` operator in the standard library calls `.into()` on the `Error` type before bubbling it up, allowing for natural conversions to other error types.  `thiserror` supports marking such errors as `#[from]` (implying `#[source]`) to generate the necessary `From<>` implementation to facilitate automatic conversion from any specified `Error` type to the corresponding enum variant in our `thiserror` enums.


ramosbugs:
Thanks!

PR #245, opened 2024-01-15
Fix comments about csrf_state

### LINKED ISSUE #208: Comments in examples mention csrf_state which does not exist in the code
https://docs.rs/oauth2/4.3.0/oauth2/index.html#example-synchronous-blocking-api

```
// Once the user has been redirected to the redirect URL, you'll have access to the
// authorization code. For security reasons, your code should verify that the `state`
// parameter returned by the server matches `csrf_state`.
```

Presumably this is meant to refer to csrf_token which is set in the code above that comment?

The same issue occurs several times in later examples.

ikehz:
Clarify that the `state` parameter should be compared to the `csrf_token.secret()`.

Fixes #208.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -131,7 +131,7 @@
 //!
 //! // Once the user has been redirected to the redirect URL, you'll have access to the
 //! // authorization code. For security reasons, your code should verify that the `state`
-//! // parameter returned by the server matches `csrf_state`.
+//! // parameter returned by the server matches `csrf_token`.
 //!
 //! // Now you can trade it for an access token.
 //! let token_result =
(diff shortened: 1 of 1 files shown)


ramosbugs:
Hey, thanks for fixing the variable name! I would prefer not to add `.secret()` as it suggests users should directly compare the secret as a string using `==`, which would create a [timing side-channel](https://en.wikipedia.org/wiki/Timing_attack) vulnerability.

To avoid introducing the vulnerability, the comparison must be done in constant time, either using a constant-time crate like [`constant_time_eq`](https://crates.io/crates/constant_time_eq) (which could break if a future compiler version decides to be overly smart about its optimizations), or by first computing a cryptographically-secure hash (e.g., SHA-256) of both values and then comparing the hashes using `==` (see #232).

The `timing-resistant-secret-traits` feature flag adds a safe (but comparatively expensive) `PartialEq` implementation to the secret types. Timing side-channels are why `PartialEq` is not auto-derived for this crate's secret types, and the lack of `PartialEq` is intended to prompt users to think more carefully about these comparisons.

I would suggest either only fixing the variable name, or adding a separate section about securely comparing secrets such as the CSRF state, and then referencing it in each example.

ramosbugs:
Thanks!

Issue #237, opened 2023-11-25
Upgrade to http 1.0


jonhoo:
[hyper has reached 1.0](https://seanmonstar.com/blog/hyper-v1/), and with it the [`http` crate did as well](https://github.com/hyperium/http/releases/tag/v1.0.0). Since `http` is part of the public API for `oauth2` (e.g., via the `pub use http` re-export), it'd be nice if `oauth2` followed suit!

I took a stab at doing it myself, and it doesn't look _too_ painful, except for the fact that there's not yet a version of `reqwest` that uses http 1.0 (https://github.com/seanmonstar/reqwest/issues/2039). Which means this is blocked for the time being, but I figured it'd still be useful to have a tracking issue.

ramosbugs:
sounds good! let's do this once `reqwest` supports 1.0, and that will require a major version bump of this crate and the downstream `openidconnect` crate.

MarijnS95:
> I took a stab at doing it myself, and it doesn't look too painful,

How much of this have you got done already? Per #236 there's now consent to see if te `Http{Request,Response}` can be replaced with those from the `http` crate. Indeed, the 1.0 release seems to mainly solidify the API rather than introduce breaking changes.

(I came around this crate for the improved `http`<->`ureq` conversions that sparked #236, and am also interested in bumping to `http 1.0.0` to be able to utilize that).

jonhoo:
I stopped fairly early on when I realized that we'd need a release of `reqwest` that also uses `http 1.0` first. That's not because of the public API, but rather just for type compatibility.

jonhoo:
Also worth linking this, which it took me a bit to find: https://hyper.rs/guides/1/upgrading/

ttys3:
for `reqwest`, we need wait upstream https://github.com/seanmonstar/reqwest/issues/2039

ramosbugs:
I've started merging breaking changes for the next major release (5.0) of this crate into `main`. Hopefully this change will be done soon enough to make it into 5.0 (no concrete timeframe).

nicolaspernoud:
Would it be possible to make a beta 5.0 branch depending on 
```toml
reqwest = { git = "https://github.com/seanmonstar/reqwest.git", branch = "hyper-v1" }
```
So that the users of this crate work on their migration to be ready for the reqwest release ?

seanpianka:
It's been merged! https://github.com/seanmonstar/reqwest/issues/2039

[...]

PR #252, opened 2024-02-20
feat(logging): instrument tracing into http clients

seanpianka:
This commit adds logging for consuming crates which enable and opt-in to tracing. These logs are configurable via the usual means, but can be tuned if other details are necessary or should be excluded.

Perhaps some of the request and response details should be omitted, however I have had difficulty troubleshooting my SSO integrations as the oauth2 library only returns "Server returned an error response" rather than the actual response details such as `"invalid_grant"` with related details. The HTTP response is printed out in a Vec<u8> and has to be converted to ASCII manually, but this at least opens the possibility for further debugging in-code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,8 @@ ureq = { version = "2", optional = true }
 url = { version = "2.1", features = ["serde"] }
 chrono = { version = "0.4", default-features = false, features = ["clock", "serde", "std", "wasmbind"] }
 serde_path_to_error = "0.1.2"
+tracing = "0.1"
+tracing-subscriber = { version = "0.3", features = ["env-filter", "parking_lot"] }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 getrandom = { version = "0.2", features = ["js"] }
@@ -48,3 +50,4 @@ uuid = { version = "0.8", features = ["v4"] }
 anyhow = "1.0"
 tokio = { version = "1.0", features = ["full"] }
 async-std = "1.6.3"
+test-log = { version = "0.2.8", default-features = false, features = ["trace"] }
--- a/src/curl.rs
+++ b/src/curl.rs
@@ -26,6 +26,7 @@ pub enum Error {
 ///
 /// Synchronous HTTP client.
 ///
+#[tracing::instrument]
 pub fn http_client(request: HttpRequest) -> Result<HttpResponse, Error> {
     let mut easy = Easy::new();
     easy.url(&request.url.to_string()[..])?;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1318,6 +1318,7 @@ where
     ///
     /// Synchronously sends the request to the authorization server and awaits a response.
     ///
+    #[tracing::instrument(err, skip(self, http_client))]
     pub fn request<F, RE>(self, http_client: F) -> Result<TR, RequestTokenError<RE, TE>>
     where
         F: FnOnce(HttpRequest) -> Result<HttpResponse, RE>,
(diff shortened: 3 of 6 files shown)


ramosbugs:
I'm open to considering trace instrumentation, but I think it should be behind a non-default feature flag to avoid adding more required dependencies, which some users are sensitive to. We also need to be careful about which fields will be logged; access/refresh/CSRF tokens and any other secrets must never be logged by this instrumentation (if users want secrets to be logged, they'll need to do it explicitly within their code).

> I have had difficulty troubleshooting my SSO integrations as the oauth2 library only returns "Server returned an error response" rather than the actual response details such as `"invalid_grant"` with related details.

The [`RequestTokenError`](https://docs.rs/oauth2/latest/oauth2/enum.RequestTokenError.html) variants are public, so users can pretty easily `match` on `RequestTokenError::ServerResponse` in their error handling code to inspect the contents of the server's error response. That response is a generic type for extensibility (i.e., some servers might return additional fields that users can access by using custom types instead of `StandardErrorResponse`), which is why its fields aren't printed as part of `RequestTokenError`'s `Display` trait implementation.

I'd be open to adding some new trait methods to [`ErrorResponse`](https://docs.rs/oauth2/latest/oauth2/trait.ErrorResponse.html) that would allow `RequestTokenError`'s `Display` implementation to access the `error`/`error_description`/`error_uri` fields for better output, but what would be a breaking change. Still, it's probably worth doing since that would help users debug the majority of errors they're likely to run into. I'll see if I can make this change soon.

> The HTTP response is printed out in a Vec and has to be converted to ASCII manually, but this at least opens the possibility for further debugging in-code.

Are you referring to the `Debug` trait's output for [`RequestTokenError::Parse`](https://docs.rs/oauth2/latest/oauth2/enum.RequestTokenError.html)? That's represented as a byte sequence because the UTF-8 conversion may have failed to parse, and all Rust strings must be valid UTF-8. Users are free to `match` on the error variants in their own code and convert the output to any format they find useful. Even if it were a `String`, it wouldn't be safe to print as part of the `Display` implementation (the typical trait used for printing errors returned by libraries) because it may contain secrets.

For debugging parse errors (or really, any errors), I highly recommend walking over the [`Error::source`](https://doc.rust-lang.org/nightly/core/error/trait.Error.html#method.source) chain to access each underlying error, which is a conventional error handling approach in Rust. Each error represents an exception at one layer of abstraction, and the `source` method returns the underlying cause from next layer of abstraction down (serde, I/O errors, etc.). See [example](https://github.com/ramosbugs/openidconnect-rs/blob/414eb35419f84ab370c0d9acc3af72d72b67462c/examples/google.rs#L38-L47). I'd recommend something like this when printing any errors in Rust. With this approach, you end up with more useful output such as:
```
ERROR: Failed to parse server response
    caused by: expires_in: invalid type: string "3599", expected u64 at line 1 column 42
    caused by: invalid type: string "3599", expected u64 at line 1 column 42
```

ramosbugs:
I made some improvements to the `RequestTokenError::ServerResponse` output in 96c6f9b17b5fdea98a6a7b84bec8e420671342eb. Feel free to test out the latest version of the `main` branch and let me know if that resolves some of your issues. This will be part of the next major release, which will probably wait for some other pending breaking changes to be merged (e.g., #237)

seanpianka:
I appreciate your thoughtful responses so quickly after I opened this PR. Were you to add tracing, it certainly makes sense to gate it with a feature flag if you know there's performance sensitive consumers of this package.

It hadn't occurred to me to match on the error type until you mentioned it, but I see that there's considerably more detail available in doing so. The AWS SDK has this same pattern, and your comments helped me debug my code elsewhere with your explanation of this pattern!

I've not used the chaining of an Error's source for my debugging much either, that's another thing I'll have to explore and understand. Thanks your example and all your help, I'll leave the proper implementation of this instrumentation to you 🙂

PR #256, opened 2024-03-06
Fix: add Send + Sync to AsyncHttpClient

LorenzoLeonardo:
Fix the error of Future not send and not safe to pass across threads.


ramosbugs:
Going to close this since, as the CI failures show, the `reqwest` client's futures aren't `Send`, which means we can't have that as a required trait bound.

LorenzoLeonardo:
Ok, I understand. Thanks very much.

PR #263, opened 2024-03-29
Derive clone for EndpointState types

avtrujillo:
Currently,  Client cannot properly derive clone, because none of the EndpointState types implement clone. For example, trying to use openidconnect's CoreClient in a context that requires it to be Clone produces the following error:

> the trait bound `openidconnect::EndpointSet: Clone` is not satisfied
> required for `Client<EmptyAdditionalClaims, ..., ..., ..., ..., ..., ..., ..., ..., ..., ..., ..., ..., ..., ..., ..., ..., ..., ..., ..., ...>` to implement `Clone`

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -35,6 +35,7 @@ impl private::EndpointStateSealed for EndpointSet {}
 
 /// [Typestate](https://cliffle.com/blog/rust-typestate/) indicating that an endpoint may have been
 /// set and can be used via fallible methods.
+#[derive(Clone)]
 pub struct EndpointMaybeSet;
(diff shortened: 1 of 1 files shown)


ramosbugs:
Thanks for the PR! This is an another unfortunate example of https://github.com/rust-lang/rust/issues/26925.

The same issue applies to `Client`'s derived `Debug` implementation. Would you mind adding `Debug` to those types as well?

avtrujillo:
> Would you mind adding `Debug` to those types as well?

Added. Thanks for the quick response!

ramosbugs:
Thanks!

PR #270, opened 2024-05-03
feat: inline format args

CosminPerRam:
Since the MSRV of 5.x will be `1.65`, [inline format args](https://github.com/rust-lang/rust/pull/90473/) were added in `1.58`, this PR replaces every possible encounter of this.
This PR comes with better code readability (marginally and only in some occasions at least) and no disadvantages.


ramosbugs:
Thanks!

PR #275, opened 2024-05-11
Use `impl Into<String>` in new type constructors

gibbz00:
Also see: https://github.com/ramosbugs/openidconnect-rs/issues/168

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -156,8 +156,8 @@ macro_rules! new_secret_type {
             $($item)*
 
             #[doc = $new_doc]
-            pub fn new(s: $type) -> Self {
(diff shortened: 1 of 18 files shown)


src/types.rs
```diff
@@ -86,8 +86,8 @@ macro_rules! new_type {
             $($item)*
 
             #[doc = $new_doc]
-            pub const fn new(s: $type) -> Self {
-                $name(s)
+            pub fn new(s: impl Into<$type>) -> Self {
```
ramosbugs, line 89: I'm realizing that having `const` here probably didn't add much value since `String`s can't be const initialized 👍

PR #279, opened 2024-06-28
Implement `schemars::JsonSchema` for non-secret new types.

gibbz00:
Placed behind a `schemars` feature flag.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -77,7 +77,8 @@ macro_rules! new_type {
         }
     ) => {
         $(#[$attr])*
-        #[derive(Clone, Debug, PartialEq)]
+        #[derive(Clone, Debug, PartialEq, Eq, Hash, ::serde::Serialize, ::serde::Deserialize)]
+        #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
         pub struct $name(
             $(#[$type_attr])*
             $type
@@ -344,13 +345,29 @@ macro_rules! new_url_type {
             }
         }
         impl Eq for $name {}
+
+
+        #[cfg(feature = "schemars")]
+        impl schemars::JsonSchema for $name {
+            fn schema_name() -> String {
+                stringify!($name).to_owned()
+            }
+
+            fn schema_id() -> std::borrow::Cow<'static, str> {
+                std::borrow::Cow::Borrowed(concat!("oauth2::", stringify!($name)))
+            }
+
+            fn json_schema(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
+                // HELP(gibbz00): do we want to generate the schema for a URL or a String?
+                gen.subschema_for::<String>()
+            }
+        }
     };
 }
 
 new_type![
     /// Client identifier issued to the client during the registration process described by
     /// [Section 2.2](https://tools.ietf.org/html/rfc6749#section-2.2).
-    #[derive(Deserialize, Serialize, Eq, Hash)]
     ClientId(String)
 ];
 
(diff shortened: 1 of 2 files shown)


src/types.rs
```diff
@@ -77,7 +77,8 @@ macro_rules! new_type {
         }
     ) => {
         $(#[$attr])*
-        #[derive(Clone, Debug, PartialEq)]
+        #[derive(Clone, Debug, PartialEq, Eq, Hash, ::serde::Serialize, ::serde::Deserialize)]
```
ramosbugs, line 80: I think there was briefly a time when not all of the new types had these traits derived, but that no longer seems to be the case, so 👍

src/types.rs
```diff
@@ -344,13 +345,29 @@ macro_rules! new_url_type {
...
+            fn schema_name() -> String {
+                stringify!($name).to_owned()
+            }
+
+            fn schema_id() -> std::borrow::Cow<'static, str> {
+                std::borrow::Cow::Borrowed(concat!("oauth2::", stringify!($name)))
+            }
+
+            fn json_schema(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
+                // HELP(gibbz00): do we want to generate the schema for a URL or a String?
```
ramosbugs, line 361: I can see this question being relevant for the `openidconnect` PR, but in this crate they're all plain strings

src/types.rs
```diff
@@ -614,7 +627,7 @@ new_secret_type![
 
 #[cfg(test)]
 mod tests {
-    use crate::{ClientSecret, CsrfToken, PkceCodeChallenge, PkceCodeVerifier};
+    use super::*;
```
ramosbugs, line 630: I'd prefer to avoid wildcard imports except from `prelude` or similar modules since they can make dependencies harder to see and grep for

src/types.rs
```diff
@@ -662,4 +675,39 @@ mod tests {
             "E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM",
         );
     }
+
+    #[cfg(feature = "schemars")]
+    mod json_schema {
+        use schemars::schema_for;
+        use serde_json::json;
+
+        use super::*;
```
ramosbugs, line 684: same

src/types.rs
```diff
@@ -662,4 +675,39 @@ mod tests {
...
+
+        #[test]
+        fn generates_new_type_json_schema() {
```
ramosbugs, line 687: nice

ramosbugs (commented):
I wasn't familiar with `schemars`, but it seems popular enough that this is worth merging. Still, I wonder if it's worth waiting for their 1.0 release (https://github.com/GREsau/schemars/pull/290) since bumping the `schemars` dependency would be a breaking change. Though, as long as `schemars` 1.0 is API-compatible (for the subset of APIs used by this crate), we could also just change `Cargo.toml` to `schemars = { version = "0.8, 1", optional = true }` once 1.0 is released without introducing a breaking change. Thoughts?

src/types.rs
gibbz00: Oki, removing it.

gibbz00:
Hey again, 

Apologies for the late reply, got a lot going on at work...
 
>  Still, I wonder if it's worth waiting for their 1.0 release (GREsau/schemars#290) since bumping the schemars dependency would be a breaking change.

Good point. I think that the safest and easiest way forward it to simply wait for the 1.0 release. It seems to be just around the corner, and I'm not in too much of a hurry 😊

Issue #260, opened 2024-03-09
(Beta version) : reference to client causes error with axum


nicolaspernoud:
Hello,

When using  oauth_client.exchange_code(AuthorizationCode::new(query.code.clone())) .request_async(&client) in an axum handler, I get the following error : 
- the trait `std::marker::Send` is not implemented for `dyn Future<Output = Result<StandardTokenResponse<EmptyExtraTokenFields, BasicTokenType>, ...>>`
- future is not `Send` as it awaits another future which is not `Send`
I use a custom client but it is implemented like the Reqwest one is.
It would seem that taking a reference to the client (which is cheap to clone anyway and should be cloned and moved is the problem).

ramosbugs:
Hey, @nicolaspernoud,

Was this working 4.x and isn't in the 5.x alpha, or is this net new code? Something similar came up in #255, but as far as I can tell this should have been an issue in 4.x too since `reqwest`'s client returns a non-`Send` future: https://docs.rs/reqwest/latest/reqwest/struct.Client.html#method.execute

> It would seem that taking a reference to the client (which is cheap to clone anyway and should be cloned and moved is the problem).

I'm not sure what you mean here. I think the issue is related to the client returning a non-`Send` future, not passing a reference vs. an owned client (not all of which may be cheap to clone).

Is there something about Axum that requires the async code to be `Send`?

LorenzoLeonardo:
Are you usng #async_trait? You try putting #async_trait(?Send)

ramosbugs:
I've been digging a bit more, and somehow the future returned by `client.exchange_code(code).request_async(oauth2::reqwest::async_http_client)` in 4.4.2 is in fact `Send` despite `reqwest::Client::execute`'s return type not explicitly listing it in its trait bounds 🤔

I'm working on adding `Send` and `Sync` bounds to all of the futures. if that compiles, I'll go ahead and merge it, which seems worth it even though it will require all custom HTTP clients to also return `Send + Sync` futures.

LorenzoLeonardo:
Are you usng #async_trait? You try putting #async_trait(?Send)

@ramosbugs so I can now use again the #[async_trait] the one without the ?Send?

ramosbugs:
[...]

Issue #255, opened 2024-03-06
Future is not Send


LorenzoLeonardo:
![image](https://github.com/ramosbugs/oauth2-rs/assets/97872577/c3f6aecb-9054-4d74-a555-2f8f1ca432d1)

I have encountered this error, I created a trait in my application so that it would be mockable during the test, but I encountered this error. Let me know what I lack.

LorenzoLeonardo:
My goal is to create a mockable AsyncHttpClient and a production AsyncHttpClient

LorenzoLeonardo:
Suggesting to put + Send here

```
impl<'c> AsyncHttpClient<'c> for Curl {
    type Error = Error<Collector>;

    fn call(
        &'c self,
        request: HttpRequest,
    ) -> std::pin::Pin<Box<dyn std::future::Future<Output = Result<HttpResponse, Self::Error>> + 'c + Send>>
    {
        Box::pin(self.send(request))
    }
}
```

LorenzoLeonardo:
PR -> https://github.com/ramosbugs/oauth2-rs/pull/256

ramosbugs:
Hmm this is tricky because the future `reqwest` returns isn't `Send` or `Sync`: https://docs.rs/reqwest/latest/reqwest/struct.Client.html#method.execute.

Note that the 4.x interface also didn't have `Send` or `Sync` bounds on the returned future (`F: Future<Output = Result<HttpResponse, RE>>`): https://docs.rs/oauth2/latest/oauth2/struct.DeviceAuthorizationRequest.html#method.request_async

It looks like `async_trait` adds `Send` bounds by default: https://docs.rs/async-trait/latest/async_trait/#non-threadsafe-futures

Does using `#[async_trait(?Send)]` on your `DeviceCodeFlowTrait` as described [here](https://docs.rs/async-trait/latest/async_trait/#non-threadsafe-futures) work?

LorenzoLeonardo:
Hello David

It is now working using #[async_trait(?Send)]. Thanks

PR #280, opened 2024-07-23
Make `AsyncHttpClient::Future` `Send` on non-wasm archs

bnjbvr:
This is useful for adding some `Send` bounds on other methods in the dependent `openidconnect-rs` crate ([here](https://github.com/poljar/openidconnect-rs/commit/c7e1dc31b83dd7559125984bfd36b9c0f191585e)). Since there was a `cfg` guard for wasm in the reqwest_client implementation, I also included it here, so that wasm doesn't require the bound itself.

### DIFF
--- a/src/endpoint.rs
+++ b/src/endpoint.rs
@@ -25,7 +25,12 @@ pub trait AsyncHttpClient<'c> {
     type Error: Error + 'static;
 
     /// Future type returned by HTTP client.
+    #[cfg(target_arch = "wasm32")]
     type Future: Future<Output = Result<HttpResponse, Self::Error>> + 'c;
+
+    /// Future type returned by HTTP client.
+    #[cfg(not(target_arch = "wasm32"))]
+    type Future: Future<Output = Result<HttpResponse, Self::Error>> + Send + 'c;
 
     /// Perform a single HTTP request.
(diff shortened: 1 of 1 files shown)


ramosbugs:
Hey @bnjbvr, I'm confused why the `Send` bound is needed either on the trait's associated type or on the methods in `openidconnect`.

Note that even if the bounds are needed in `openidconnect`, Rust 1.79 stabilized [bounds on associated types](https://blog.rust-lang.org/2024/06/13/Rust-1.79.0.html#bounds-in-associated-type-position), which should avoid the need to add the bound here (at the cost of incrementing the MSRV in the other crate).

I believe 6e583bd03203e42ef712fc90edb57cf5a389f9b7 should have fixed the issues around `Send` bounds. When using a `Send` HTTP client, the compiler should automatically infer `Send` for the returned future. See related discussions in #260, #255, and ramosbugs/openidconnect-rs#171.

ramosbugs:
If possible, I'd like to continue supporting both `Send` and non-`Send` HTTP clients.

bnjbvr:
Ah indeed, thanks for the push back, I hadn't retried with the latest commit of the crate, and turns out the `Send` bound isn't required anymore. Closing this.

PR #281, opened 2024-08-06
feat: enable user to disable token exchange response body

PixelWeaver:
Hi there, great lib, thanks for your work on this! I've been using a fork of it for a bit it's been working really well!

Some APIs (e.g. [Officient](https://apidocs.officient.io/)) do not respond to the token exchange request with one of the recognized Content-Type headers, causing `check_response_body` to return `false`.

Adding the `disable_check_response_body` function to the builder of `CodeTokenRequest` would offer a trifle more flexibility to users of this library without altering the default behavior.

### DIFF
--- a/src/devicecode.rs
+++ b/src/devicecode.rs
@@ -178,7 +178,7 @@ where
         C: SyncHttpClient,
         EF: ExtraDeviceAuthorizationFields,
     {
-        endpoint_response(http_client.call(self.prepare_request()?)?)
+        endpoint_response(http_client.call(self.prepare_request()?)?, true)
     }
 
     /// Asynchronously sends the request to the authorization server and returns a Future.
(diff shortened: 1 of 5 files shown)


ramosbugs:
What `Content-Type` header value is returned? The [spec](https://datatracker.ietf.org/doc/html/rfc6749#section-5.1) is pretty clear that the response body needs to be JSON:
> The parameters are included in the entity-body of the HTTP response
   using the "application/json" media type as defined by [[RFC4627](https://datatracker.ietf.org/doc/html/rfc4627)]

This library treats missing `Content-Type` headers as if they're JSON, so if `check_response_body()` is returning an error it means that either the `Content-Type` response header is explicitly set to something other than `application/json`, or the response body is missing entirely. Neither of those cases are compliant with the spec, so I don't think it makes sense to add explicit support to this library.

My suggestion, as for most other deviations from the spec, is to define a custom HTTP client as a shim (passed to `request()`/`request_json()`) that converts the non-compliant response into a compliant one before returning it to this library. For example, that could mean setting the `Content-Type` header to `application/json` in the HTTP response.

PR #306, opened 2025-03-13
feat(deps): update rand

CosminPerRam:
EDIT: This PR originally went to update `getrandom` too, but that includes a wasm breaking change and can be done in a major version only.

Updates `rand` from `0.8` to `0.9` ~and `getrandom` from `0.2` to `0.3`~.

The `rand` update included some deprecations, I've replaced them.
~The `getrandom` update renamed the `js` feature to `wasm_js`.~

I'd like this to be merged as I currently use other crates that are on the newer versions (and this one adds duplicated deps).

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[target.wasm32-unknown-unknown]
+rustflags = ['--cfg', 'getrandom_backend="wasm_js"']
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,0 @@
-[target.wasm32-unknown-unknown]
-rustflags = ['--cfg', 'getrandom_backend="wasm_js"']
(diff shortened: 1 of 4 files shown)


CosminPerRam:
Hopefully I've correctly updated `Cargo-1.65.lock`, if not, please mention to revert.

CosminPerRam:
@ramosbugs bump, any blockers for this to be merged?

ramosbugs:
does this require every WASM build to set `rustflags = ['--cfg', 'getrandom_backend="wasm_js"']`? if so, I think this is a breaking change that will have to wait for the next major release

CosminPerRam:
Unfortunately yes, I'll remove the `getrandom` dependency update from this PR for now.

CosminPerRam:
Apparently neither can `rand` be updated to `0.9` as it depends on `rand_core 0.9` which itself depends on `getrandom 0.3`, which requires the flags, closing this for now.

PR #307, opened 2025-03-14
Update to ureq 3

softdevca:
Simple update to ureq version 3, no changes required except for `Cargo.toml`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,7 @@ reqwest = { version = "0.12", optional = true, default-features = false }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 sha2 = "0.10"
-ureq = { version = "2", optional = true }
+ureq = { version = "3", optional = true }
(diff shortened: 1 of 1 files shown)


ramosbugs:
since `ureq` is re-exported as part of the public interface when the `ureq` feature is enabled, this is a breaking change that would need to wait for the next major release (not planned any time soon). however, I'd be willing to accept the change behind a separate feature flag (e.g., `ureq-3`) that preserves the current version when the `ureq` feature is used.

SO9010:
Hello, 

Would it be possible to get the re-exporting done through a flag? This would be useful for the project I'm part of to reduce dependencies. 

Many thanks, 
SO9010

hanyuone:
@SO9010 did you end up finding out what needed to be done for this PR to work?

SO9010:
It is quite a bit of work, ureq-3 is completely different... you just need to go through and reimplement all of usages. It's not hard just pretty damn verbose.

algesten:
I had a go at this in #324

PR #310, opened 2025-04-21
Fix: DeviceAccessTokenRequest: Handle response.error even if the server returned 200...

pfnsec:
in the device flow, we poll the server continuously for a device token and, depending on the server's response, slow down, or repoll, or we try and parse the response body as a StandardTokenResponse.

https://github.com/ramosbugs/oauth2-rs/blob/096b286b4faac77eb22276f62535f9f53e9d8610/src/devicecode.rs#L390

However, unless this is an "I'm holding it wrong" problem, I think some wires have gotten tangled.

When we run the included example, `github_devicecode_async`, `exchange_device_access_token(...)` throws the following error on parsing the server's response:

`RequestTokenError::Parse(Error { path: Path { segments: [] }, original: Error("missing field `access_token`",`

This is from [endpoint_response](https://github.com/ramosbugs/oauth2-rs/blob/096b286b4faac77eb22276f62535f9f53e9d8610/src/endpoint.rs#L170), where the generic argument attempts to parse a StandardTokenResponse. But what is it actually receiving?

We get a response status of 200, and in the response body, we get:
```
{
    "error":"authorization_pending",
    "error_description":"The authorization request is still 
    pending.","error_uri":"https://docs.github.com/developers/apps/authorizing-oauth-apps#error-codes-for-the-device- 
    flow"
}
```

Since this isn't the form of a StandardTokenResponse, the deserialize fails. However, this case should be caught earlier in the response-processing logic, and the error should be parsed as one of the internal DeviceCodeErrorTypes and handled accordingly. This pull request implements that new logic.

### DIFF
--- a/examples/github_devicecode_async.rs
+++ b/examples/github_devicecode_async.rs
@@ -1,0 +1,72 @@
+//!
+//! This example showcases the Github OAuth2 device flow process for requesting access to the user's public and private repos.
+//!
+//! Before running it, you'll need to generate your own Github OAuth2 credentials.
+//!
+//! In order to run the example, call:
+//!
+//! ```sh
+//! GITHUB_CLIENT_ID=xxx cargo run --example github_devicecode_async
+//! ```
+//!
+//! ...and follow the instructions.
+//!
+//! Note that this example does not require a client secret, making it useful 
+//! for building local applications that require Github auth.
+//!
+
+use oauth2::basic::BasicClient;
+use oauth2::{
+    reqwest, DeviceAuthorizationResponse, DeviceAuthorizationUrl,
+    EmptyExtraDeviceAuthorizationFields,
+};
+use oauth2::{
+    AuthUrl, ClientId, Scope,
+    TokenUrl,
+};
+use url::Url;
+
+use std::env;
+
+#[tokio::main]
+async fn main() {
+    let github_client_id = ClientId::new(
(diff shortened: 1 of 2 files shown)


ramosbugs:
> When we run the included example, `github_devicecode_async`

please link to this example, because I don't see it [in the repo](https://github.com/ramosbugs/oauth2-rs/tree/main/examples). did you use an LLM to generate that?

the [spec](https://datatracker.ietf.org/doc/html/rfc8628#section-3.5) is unambiguous here:
> If the user has approved the grant, the token endpoint responds with
   a success response defined in [Section 5.1 of [RFC6749]](https://datatracker.ietf.org/doc/html/rfc6749#section-5.1); otherwise, it
   responds with an error, as defined in [Section 5.2 of [RFC6749]](https://datatracker.ietf.org/doc/html/rfc6749#section-5.2).

those references say that `200 OK` is when the server issues an access token vs. `400 Bad Request` for an error response.

my typical recommendation for dealing with non-compliant providers is to pass a custom HTTP client as a shim that converts the response to an RFC-compliant one before returning it for this crate to process.

pfnsec:
I'm not surprised that a Github API appears to be violating the spec, but wouldn't it make this library more useful to gracefully handle this case, instead of putting the onus on the user to figure this out and construct their own custom client? 

The bug in question was difficult to discover and wasted several hours of engineering work. Perhaps we can at least put a notice in the README to save future users some time:

## Warning: Device-flow does not worth with non-compliant providers (Github)!

PR #312, opened 2025-04-28
Implement `FromStr` for String newtypes

mhutter:
Allows for easier deserialization of configuration.

Relates: https://github.com/ramosbugs/openidconnect-rs/issues/202


ramosbugs:
See https://github.com/ramosbugs/openidconnect-rs/issues/202#issuecomment-3042902922

PR #319, opened 2025-05-21
Update rand to 0.9

### LINKED ISSUE #318: Update to rand 0.9
Using the older verison of rand is currently causing issues with other depdencies which have updated to use the newest version of rand.

JonathanWoollett-Light:
Closes https://github.com/ramosbugs/oauth2-rs/issues/318

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,7 @@ required-features = ["reqwest-blocking"]
 base64 = ">= 0.21, <0.23"
 thiserror = "1.0"
 http = "1.0"
-rand = "0.8"
+rand = "0.9"
 reqwest = { version = "0.12", optional = true, default-features = false }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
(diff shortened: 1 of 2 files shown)


hanyuone:
From `rand 0.10` onwards, `getrandom 0.4` is used, which no longer requires any extra `rustflags`, thus not leading to a breaking change for a major version (the issue seen in #306).

@ramosbugs would a PR that bumps the `rand` version to `0.10` be accepted?

CosminPerRam:
@hanyuone probably not (unless with a new major), the MSRV for `0.10` is `1.85.0`, whereas we are on `1.65.0`

hanyuone:
@CosminPerRam I see - although could #306 be merged now, since `getrandom v0.3.4` uses the `wasm_js` backend by default, thus removing the need for setting `rustflags`? See the changelog for `getrandom` [here](https://github.com/rust-random/getrandom/blob/master/CHANGELOG.md#034---2025-10-14).

lovasoa:
Hi ! I would love to see this merged. I currently have three different versions of rand built into [sqlpage](https://sql-page.com/) ! 

I think I'm not the only one who wouldn't mind a new major version being released.

xtqqczze:
@JonathanWoollett-Light this branch has conflicts

PR #324, opened 2025-09-10
Upgrade to ureq3

algesten:


src/ureq_client.rs
```diff
@@ -1,60 +1,17 @@
...
-        {
-            builder = builder.header(CONTENT_TYPE, content_type);
-        }
+        let response = self.run(request).map_err(Box::new)?;
 
-        let mut body = Vec::new();
-        response.into_reader().read_to_end(&mut body)?;
+        let (parts, body) = response.into_parts();
+        let mut body_vec = Vec::new();
+        body.into_reader().read_to_end(&mut body_vec)?;
```
algesten, line 13: I think this should be changed to 

```rs
let body_vec = body.read_to_vec()?;
```

That function has a 10MB cap, which is a safety mechanism to avoid a misbehaving remote server causing problems.

I struggled to find a good `Error` variant to use though.

Issue #298, opened 2025-01-19
Authorization Server Metadata and Dynamic Client Registration support


zecakeh:
Hi, I wondered if there was any interest to add support for those two OAuth 2.0 RFCs:

1. Authorization Server Metadata ([RFC8414](https://datatracker.ietf.org/doc/html/rfc8414)): The main work would probably be to have a deserializable type for the metadata, since there is not mandatory endpoint to use to request it. It could even be used to build the `Client`, like [`Client::from_provider_metadata()`](https://docs.rs/openidconnect/latest/openidconnect/struct.Client.html#method.from_provider_metadata) in openidconnect. It would be very close to the [`ProviderMetadata`]() from openidconnect, except that the required fields are different, like `jwks_uri` is not required for OAuth.
2. Dynamic Client Registration ([RFC7591](https://datatracker.ietf.org/doc/html/rfc7591)): this is one is a proper request with the `registration_uri` endpoint but the main work will probably be the type for the client metadata too. Here the [`ClientMetadata`](https://docs.rs/openidconnect/latest/openidconnect/registration/struct.ClientMetadata.html) looks like it could fit. 
 
Since openidconnect depends on this crate, maybe there can be a way to create the metadata types here and reuse them in openidconnect so they can be extended if necessary?

keskalukasfri:
We're blocked by needing DCR, so we're working on a branch to get this done, in case anyone is looking at it.

PR #348, opened 2026-06-22
feat: Dynamic client registration support

keskalukasfri:
This PR adds support for DCR ([RFC 7591](https://datatracker.ietf.org/doc/html/rfc7591)). It adds request types for request/response, as well as a new "DynamicClient" type which simply wraps the request-type, made to keep types from the initialization of the client through registering it, turning it automatically into an oauth2::Client.

This PR was cowritten using AI, but it's not vibecoded and I think it's fairly reasonable, though feedback on the API surface is very much appreciated.

### DIFF
--- a/oauth2/Cargo.toml
+++ b/oauth2/Cargo.toml
@@ -47,6 +47,10 @@ required-features = ["reqwest-blocking"]
 name = "wunderlist"
 required-features = ["reqwest-blocking"]
 
+[[example]]
+name = "dynamic_client_registration"
+required-features = ["reqwest-blocking"]
+
 [dependencies]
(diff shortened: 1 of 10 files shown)


keskalukasfri:
This partially closes #298 .

keskalukasfri:
I realized a bit late that I did not leave any way to get out the client secret post registration. I changed register* to now return a tuple with the response, but I'm not quite happy with this. The reason for not simply returning the response (and letting users call into_client) is to not erase the types, or have to create *another* wrapper to simply keep type data, but this design isn't ideal. I welcome any suggestions for a better API design, but this will suffice for our own work for now.

